PR #56, opened 2015-07-02
Add a file to generate python swig bindings

haudren:
It adds the files, generates a custom target and compiles it
It needs more work to support more languages and less configuration.

For now, you need to specifiy "by hand" the python version... It also relies on a "hack" to find all previously defined includes.
Basic usage is as follows :
```cmake
SET(Python_ADDITIONAL_VERSIONS 2.7)
FINDPYTHON()
INCLUDE_DIRECTORIES(SYSTEM ${PYTHON_INCLUDE_DIRS})
ADD_SWIG_PYTHON_BINDING(${source} ${directory})
GENERATE_SWIG_BINDINGS()
BUILD_SWIG_BINDINGS(${DEPENDENCIES})
```


florent-lamiraux:
+1: I am not a big fan of swig, but adding a separate file does not hurt.

bchretien:
:+1: 

> For now, you need to specifiy "by hand" the python version...

Could you elaborate? Is this because this only works with Python 2?

Also, a bit of doc wouldn't hurt (just a sentence explaining what the macros do at least), since people will usually look at the code to know which macro to use.

bchretien:
Also, it seems the swig module in CMake has been improved in the last CMake versions (or at least [its doc](http://www.cmake.org/cmake/help/v3.2/module/UseSWIG.html) grew). I just don't know how mature it is and if we can have as much control as with our own set of macros.

haudren:
@bchretien  : No it's because you need to specify with which python lib to link with, and this (afaik) must be set beforehand using SET(Python_ADDITIONAL_VERSIONS my_version) from python.cmake . But it is a good point, I did not even suspect that cmake had built-in support for swig, which may make this moot. After a quick look at their documentation, it does roughly the same thing and looks cleaner. I'll test soon, but if anyone feels like it before me :)

bchretien:
AFAIK, setting `Python_ADDITIONAL_VERSIONS` is useful if you want a non-default Python version, since CMake will start its lookup with the user-provided ones, then it continues looking backwards from Python 3.4. On a side note, CMake being CMake, I would not just count on that option for systems with an unusual default Python version (e.g. Arch Linux), since you end up with a mix of interpreter, libraries and include directories of different versions (or at least you did not so long ago). In that case, it is advised to do it [the old-fashioned way](https://aur4.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=roboptim-core-python-git#n51).

Having a wrapper around this feature to make it easier would make sense too, if you think that the stock macros are too much of a hassle. The minimum required CMake version may also be part of the equation (or we just assume that people finally gave their farewell to CMake 2.8).

haudren:
@olivier-stasse : I am not sure this was mergeable as it is not commented as pointed out by @bchretien . Moreover, the newer cmake swig module is now quite complete. As we do not (in the end) finally need swig bindings in our projects, I let go of this branch.

If this is meant to be kept in master, I will add at least a few comments and warnings about how this has not been extensively used and compared to default cmake swig.

olivier-stasse:
@haudren Should we then revert the merge ?

haudren:
I don't think that anyone is using swig at the moment, so yes we should revert the merge. As I don't see anyone really interested in adding SWIG bindings, the PR should probably be closed as well.

We have discussed today with @gergondet about future endeavours that may require swig, but if it is necessary, I will reopen an updated PR.

Sorry for the inconvenience!

PR #57, opened 2015-07-18
Doxygen improvements

bchretien:
* Add support for MathJax, which removes build dependencies on latex, dvips and gs.
* Bundle a minimal MathJax  (~2 MB) so that it can be installed with the doc (which can thus be read offline).
* Add support for multiple formats for images generated by dot (and use SVG as a default format to reduce the size of generated docs).


bchretien:
cc @florent-lamiraux who was recently improving Doxygen support.

PR #58, opened 2015-07-23
Add PROJECT_DEBUG_POSTFIX

gergondet:
Setting this variable has the following effects:
- Set CMAKE_DEBUG_POSTFIX to PROJECT_DEBUG_POSTFIX
- Generate a different .pc file named ${PROJECT_NAME}${PROJECT_DEBUG_POSTFIX}.pc when building in Debug mode
- Generate two .pc files for multiple build types generators (e.g. Visual Studio)

The following macros are added:
- PKG_CONFIG_APPEND_CFLAGS_DEBUG
- PKG_CONFIG_APPEND_CFLAGS_OPTIMIZED

They allow to add specific flags for the debug and optimized version of the project.

Finally, PKG_CONFIG_APPEND_LIBS was modified to apply the postfix on libraries built by the project.


bchretien:
:+1: This should not break anything for existing setups, and we have an actual project ready for further testing. @aescande could even confirm a posteriori that things also work as expected on Windows + Visual Studio.

PR #59, opened 2015-07-23
Update header and footer of doxygen

jmirabel:
* Enable the tree view to ease navigation
* Enable folding/unfolding of folder on recent version of Doxygen. For older version, this should not cause more trouble than a script file not found, thus ignored, by the browser. But it has not been tested.
* Add footer
* Add missing tags at the end of header.html

### DIFF
--- a/doxygen/footer.html
+++ b/doxygen/footer.html
@@ -1,2 +1,21 @@
+<!-- HTML footer for doxygen 1.8.6-->
+<!-- start footer part -->
+<!--BEGIN GENERATE_TREEVIEW-->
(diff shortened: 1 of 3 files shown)


bchretien:
An [extra header](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/doxygen/header-mathjax.html) was recently added for MathJax support, so I guess it could also benefit from the changes you made to `header.html`.

jmirabel:
You are right. I added my changes to header-mathjax.html

bchretien:
:+1:

PR #60, opened 2015-07-24
Topic/add pkg config alternative

gergondet:
Hi,

This is a follow-up to Issue #58 

Essentially, this allows you to specify a second PKG_CONFIG_STRING that will be used when building in debug transparently. This means that once you have defined this secondary library, e.g:
`ADD_REQUIRED_DEPENDENCY("foo >= 1.0" "foo_d >= 1.0")`

And if you use this dependency through the proper macro:
`PKG_CONFIG_USE_DEPENDENCY(myTarget foo)`

Then when building in debug mode, `myTarget` will link with the libraries provided by the `foo_d` package and in all other modes it will link with the libraries provided by the `foo` package.

If an additional string is not specified then the behavior should be similar to previous versions of jrl-cmakemodules.

Finally, PKG_CONFIG_USE_LCOMPILE_DEPENDENCY was greatly simplified using CMake >= 2.8 features (`TARGET_COMPILE_OPTIONS` and CMake generator expressions).


gergondet:
An update of this PR.

It has been tested in the real-world and works well with two exceptions:
- ADD_*_DEPENDENCY that use an alternative have to be done after those that don't. This is likely a bug in the macros that I will get to
- PKG_CONFIG_APPEND_LIBS (this is actually related to #58) has to be done in the same source directory as the corresponding ADD_LIBRARY if we want CMAKE_DEBUG_POSTFIX to be added to the pc file. I believe this is a cmake issue, a possible solution would be to emit a warning when PKG_CONFIG_APPEND is used with a non target but this would emit warning for legit use of the macro (e.g. adding lib flags for a library that does not support pkg-config)

When the first issue is fixed I think we can merge this if nobody disagrees.

gergondet:
Hi,

The first issue raised in the previous comment is fixed with the latest commits.

The second issue actually only happens (AFAIK) when one use PKG_CONFIG_APPEND_LIBS before using add_library which is simply a mistake on the part of the programmer in that case.

Finally, a small change on dependency was introduced in the last commit. The debug package provided to ADD_*_DEPENDENCY follows the requirement policy of the non-debug package except in single build generators where, if the main package is required:
- in Debug mode, the debug package is required but the main package is not
- in other modes, the main package is required but the debug package is not

This should greatly simplify the build process when one only wants to build the debug or release version.

olivier-stasse:
:+1:

pkg-config.cmake
```diff
@@ -478,23 +602,19 @@ ENDMACRO(PKG_CONFIG_APPEND_LIBS)
...
+  IF(DEFINED ${PREFIX}_DEBUG_FOUND)
+    FOREACH(FLAG ${${PREFIX}_DEBUG_CFLAGS})
+      TARGET_COMPILE_OPTIONS(${TARGET} PUBLIC $<$<CONFIG:Debug>:${FLAG}>)
+    ENDFOREACH()
+    FOREACH(FLAG ${${PREFIX}_CFLAGS})
+      TARGET_COMPILE_OPTIONS(${TARGET} PUBLIC $<$<NOT:$<CONFIG:Debug>>:${FLAG}>)
+    ENDFOREACH()
+  ELSE()
+    FOREACH(FLAG ${${PREFIX}_CFLAGS})
+      TARGET_COMPILE_OPTIONS(${TARGET} PUBLIC ${FLAG})
```
bchretien: `TARGET_COMPILE_OPTIONS` should be replaced with `SET_TARGET_PROPERTIES` if it does work with generators, else this won't work on old CMake versions, e.g. the one on Ubuntu 12.04 (used on Travis).

gergondet:
The previous commit fixes the issue raised by @bchretien and keeps the jrl-cmakemodules compataible with cmake < 2.8.12 (i.e. get rids of target_compile_options)

I am fairly confident this can be merged now.

bchretien:
:+1: This seems to be working on Travis for `roboptim-core-manifold` with `manifolds` using this feature (although the test is not done in Debug), fixing roboptim/roboptim-core-manifold#19.

PR #61, opened 2015-07-31
Rely on updated GNUInstallDirs.cmake for multi-arch support

bchretien:
cc @AdamWill

This is related to #22, the [discussion](https://github.com/jrl-umi3218/jrl-cmakemodules/commit/fba6aef9c4cddcc9728192bd17d0c1906a9b2dcd) that followed, d54df3bc05bc19f3aa56d7b7cb15fca48127861a, and roboptim/roboptim-core#101. After the `GNUInstallDirs.cmake` update, multi-arch should now be automatically handled when the prefix is set to `/usr` on multi-arch systems. `CMAKE_INSTALL_LIBDIR` can still be modified by the user to override the defaults if required (e.g. when packaging on Arch Linux, it defaults to `lib64`, but `/usr/lib64 -> /usr/lib` and the package manager refuses to install explicitly to symlinked directories).


AdamWill:
That certainly would be good for Fedora, thanks! :+1:

PR #62, opened 2015-08-17
Add a module to find QPOases solver. USE : SEARCH_FOR_QPOASES()

MaximilienNaveau:
Add a cmake module to link against QPOases a solver for quadratic optimisation problem.

### DIFF
--- a/FindqpOASES.cmake
+++ b/FindqpOASES.cmake
@@ -1,0 +1,53 @@
+#.rst:
+# FindqpOASES
+# -----------
+#
+# Try to find the qpOASES library.
+# Once done this will define the following variables::
+#
+#  qpOASES_FOUND         - System has qpOASES
+#  qpOASES_INCLUDE_DIRS  - qpOASES include directory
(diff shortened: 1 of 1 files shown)


jmirabel:
:+1:

MaximilienNaveau:
In fact the .cmake is not good enough to be merged so I will close this pull request.

bchretien:
Another possibility would be to send a patch to qpOASES developers adding pkg-config support. They already use CMake so this is fairly straightforward, but then users would need to use the latest version. AFAIK it's not packaged in many distributions yet (I'm maintaining the one for Arch, but for Ubuntu it's [in their todo list](https://projects.coin-or.org/qpOASES/ticket/25)), so this should not be much of a problem either.

MaximilienNaveau:
In fact, we used a "private cmake" folder containing the findqpoases.cmake file to not pollute the jrlcmakemodule with a not well written cmake file.

jmirabel:
For those who are interested, I have packaged qpOASES using this repository: https://github.com/jmirabel/qpOASES

PR #63, opened 2015-09-04
idl.cmake: add Python support

bchretien:
cc @gergondet 

CMake target dependencies (to generate these automatically) are still not working, since apparently the CMake support for this is only valid if everything is in the same directory....


idl.cmake
```diff
@@ -51,21 +51,81 @@ MACRO(GENERATE_IDL_FILE FILENAME DIRECTORY)
...
+#
+# Generate C++ stubs from an idl file (legacy macro).
+# An include directory can also be specified.
+#
+# FILENAME : IDL filename without the extension
+#            Can be prefixed by a path: _path/_filename
+# DIRECTORY : IDL directory
+#             The idl file being search for is: ${DIRECTORY}/${_filename}.idl
+#
+MACRO(GENERATE_IDL_FILE FILENAME DIRECTORY)
```
gergondet: IMHO, we should get rid of this macro, I don't think backward compatibility is very important for this feature:
a\ It's used by very few projects (actually the only one that comes to mind now is for the integration between the pre-python stack of tasks and the old version of OpenHRP)
b\ to be fully retro-compatible, we would need to also reproduce the variables (i.e. `ALL_IDL_STUBS`)
c\ the name of the macro was confusing in the first place

haudren:
:+1: I don't think backwards compatibility is necessary at all... 

On a side note, I would have liked to be able to do everything from a single macro but we still need to do things by hand:
- set_source_file_properties(myidlSK.cc properties generated 1) : This is unfortunately not possible in current versions of CMake "Source file properties are visible only to targets added in the same directory (CMakeLists.txt)" cf.[doc]
- We need to explictly make the desired executable/lib dependent on a target that encompasses the idl generation, i.e. via:
```
ADD_CUSTOM_TARGET(generate_idl_cpp DEPENDS ${ALL_IDL_CPP_STUBS})
ADD_CUSTOM_TARGET(generate_idl_python ALL DEPENDS ${ALL_IDL_PYTHON_STUBS})
```
And then
```
add_dependencies(mytarget generate_mc_idl_cpp)
```
I don't think there is any other way to do this (at least for now and the [FAQ] seems to confirm it as well) but I think it should be documented here even we don't come up with clever macros to solve this problem...

[doc]: http://www.cmake.org/cmake/help/v3.0/command/set_source_files_properties.html 
[FAQ]: http://www.cmake.org/Wiki/CMake_FAQ#How_can_I_add_a_dependency_to_a_source_file_which_is_generated_in_a_subdirectory.3F

bchretien:
The older function is removed, and I added a word on the build dependency problem with a link to the FAQ.

bchretien:
Any final comment before merging?

olivier-stasse:
I agree with the previous comments and prefer to use the new naming which is clearer. We may have some issues with the new scheme but I prefer to port them.

florent-lamiraux:
If possible, it would be nice to keep macro "GENERATE_IDL_FILE", but to make this macro generate an error message that tells developers how to upgrade the cmake code.

PR #64, opened 2015-09-14
PKG_CONFIG_APPEND_LIB handles absolute path properly

gergondet:
Hi,

This PR fixes PKG_CONFIG_APPEND_LIB when an absolute path is provided. This is typically the case when doing something like:

```find_package(Boost REQUIRED COMPONENTS system filesystem thread)
PKG_CONFIG_APPEND_LIBS(${Boost_LIBRARIES})```


bchretien:
:+1: Looks good to me.

Issue #68, opened 2015-09-24
Fallback to MathJax doc if LaTeX/dvips are not available


bchretien:
Right now, this is solely controlled by the user-defined variable.

There is also a dependency on perl, but I am not sure what uses it. Is this still required?

PR #70, opened 2015-09-30
Improve doc target

gergondet:
Hi,

This PR is related to #68. Namely it falls back to MathJax generation if latex, dvips and/or gs are not found.

Additonnaly, the doc target is also skipped altogether if Doxygen is not found.

I have also removed the dependency to Perl. It can be used in conjunction with Doxygen but it seems that we disable this in the default Doxygen file, hence the dependency should be introduced on a project by project basis.

Cheers


bchretien:
:+1:

Galdeano:
+1, Travis check will be faster without latex installation.

gergondet:
Closes #68

PR #71, opened 2015-10-26
Fix for ld >= 2.24.90

bchretien:
This needs to be tested on OSX + Ubuntu.

Related discussion for ROS: ros/catkin#694


pkg-config.cmake
```diff
@@ -287,6 +287,13 @@ MACRO(ADD_DEPENDENCY P_REQUIRED DOC_ONLY PKG_CONFIG_STRING PKG_CONFIG_DEBUG_STRI
     ENDIF(${P_DEBUG_REQUIRED})
   ENDIF()
 
+  # Fix for ld >= 2.24.90: -l:/some/absolute/path.so is no longer supported.
+  # See shared-library.cmake.
+  IF(UNIX AND NOT ${LD_VERSION} VERSION_LESS "2.24.90")
+    STRING(REPLACE ":/" "/" "${PREFIX}_LIBRARIES" "${${PREFIX}_LIBRARIES}")
+    STRING(REPLACE "-l:/" "/" "${PREFIX}_LDFLAGS" "${${PREFIX}_LDFLAGS}")
+  ENDIF()
```
gergondet: Please do the same with the PREFIX_DEBUG variables

shared-library.cmake
```diff
@@ -24,6 +24,19 @@ IF(WIN32)
...
-  SET(LIBINCL_ABSKW "-l:")
   SET(LIB_EXT "")
+
+  # Using -l:/some/absolute/path.so was an "undocumented ld feature, in
+  # actual fact a ld bug, that has since been fixed".
+  # This was apparently used (e.g. in ROS) because of pkg-config problems that
+  # have since been fixed.
+  # See: https://github.com/ros/catkin/issues/694#issuecomment-88323282
+  # Note: ld version on Linux can be 2.25.1 or 2.24
+  EXECUTE_PROCESS(COMMAND ${CMAKE_LINKER} -v OUTPUT_VARIABLE LD_VERSION_STR)
```
gergondet: This does not work on OSX because ld gives its output on the error output rather than the standard one, fortunately cmake can merge the two so:
EXECUTE_PROCESS(COMMAND ${CMAKE_LINKER} -v OUTPUT_VARIABLE LD_VERSION_STR ERROR_VARIABLE LD_VERSION_STR)
will work on both of these UNIX variants

shared-library.cmake
```diff
@@ -24,6 +24,19 @@ IF(WIN32)
...
+  STRING(REGEX MATCH "([0-9]+.[0-9]+.?[0-9]*)" LD_VERSION ${LD_VERSION_STR})
```
gergondet: I suggest:
([0-9]+\\.[0-9]+(\\.[0-9]+)?)
This parses the version correctly for both the GNU Linker and OSX linker (which is version 253.6 on Xcode 7.0 and works as ld > 2.24.90 w.r.t -l: support

bchretien: :+1:

shared-library.cmake
bchretien: Updated the commit with this.

pkg-config.cmake
bchretien: Fixed in the updated commit.

gergondet:
Looks good to me now. I will merge once a few more people have had some time to take a look at it.

bchretien:
Ok, no complaints so merging ;)

PR #72, opened 2015-10-26
Doxygen stylesheet

jmirabel:
### Problem
As mentioned in the Doxygen manual (http://www.stack.nl/~dimitri/doxygen/manual/customize.html#minor_tweaks_header_css), the style sheet should not be passed with `HTML_STYLESHEET` but with `HTML_EXTRA_STYLESHEET`. Tag `HTML_STYLESHEET` produces conflicts between Doxygen versions.

### Proposed solution
To ensure backward compatibility, variable `DOXYGEN_USE_TEMPLATE_CSS` is introduced. When set to `TRUE`, the default Doxygen CSS is used instead of the one provided by this module (outdated on Ubuntu 14.04).
The variable defaults to `YES` and users willing to have backward compatibility should add `SET (DOXYGEN_USE_TEMPLATE_CSS NO)` to `CmakeLists.txt`.

### Cleaner solution
A cleaner solution would be to replace in `doxygen/Doxyfile.in` this line:
```HTML_STYLESHEET        = @PROJECT_BINARY_DIR@/doc/doxygen.css```
by
```HTML_STYLESHEET        =```
and to remove `doxygen/doxygen.css`
Users could add their own CSS by setting `HTML_EXTRA_STYLESHEET` in their `Doxyfile.extra`. I do not see any easy way of ensuring backward compatibility with this solution.

### DIFF
--- a/doxygen.cmake
+++ b/doxygen.cmake
@@ -47,6 +47,12 @@ MACRO(_SETUP_PROJECT_DOCUMENTATION)
       SET(DOXYGEN_USE_MATHJAX "NO")
     ENDIF()
 
+    # HTML style sheet configuration
+    IF(NOT DEFINED DOXYGEN_USE_TEMPLATE_CSS)
+      SET(DOXYGEN_USE_TEMPLATE_CSS "NO")
+    ENDIF()
+
+
     # Teach CMake how to generate the documentation.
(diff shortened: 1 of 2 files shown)


bchretien:
:+1:

PR #73, opened 2015-10-28
Fix HEADER_INSTALL when installing headers in CMAKE_INSTALL_INCLUDEDIR

jmirabel:
file is `include/header.hpp` should be installed in `${CMAKE_INSTALL_INCLUDEDIR}/header.hpp` and not  `${CMAKE_INSTALL_INCLUDEDIR}/include/header.hpp`


bchretien:
:+1: Good catch.

PR #74, opened 2016-01-13
[shared-library] Find CMAKE_LINKER if the variable does not exist.

olivier-stasse:
There is an issue when CMAKE_LINKER does not exist (happen on travis - 12.04 LTS /precise), it makes the  [REGEXP test fails](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/shared-library.cmake#L35). This pull request fixes it by calling CMakeFindBinUtils if the variables is not set.

### DIFF
--- a/shared-library.cmake
+++ b/shared-library.cmake
@@ -32,6 +32,10 @@ ELSEIF(UNIX)
   # have since been fixed.
   # See: https://github.com/ros/catkin/issues/694#issuecomment-88323282
   # Note: ld version on Linux can be 2.25.1 or 2.24
+  IF (NOT CMAKE_LINKER)
+    INCLUDE(CMakeFindBinUtils)
+  ENDIF()
+
   EXECUTE_PROCESS(COMMAND ${CMAKE_LINKER} -v OUTPUT_VARIABLE LD_VERSION_STR ERROR_VARIABLE LD_VERSION_STR)
   STRING(REGEX MATCH "([0-9]+\\.[0-9]+(\\.[0-9]+)?)" LD_VERSION ${LD_VERSION_STR})
   IF(${LD_VERSION} VERSION_LESS "2.24.90")


bchretien:
:+1: Good catch, thanks!

olivier-stasse:
Thanks for the quick merge !

gergondet:
Hi @olivier-stasse 

Is this related to stack-of-tasks/soth#8 ?

I had a similar problem recently and I found that the issue was that project() was called in the CMakeLists.txt before the SETUP_PROJECT macro (which also calls project()) which is also the case in stack-of-tasks/soth I think the big difference is that in the macro we call project with the CXX argument and thus this fix may be hiding a bigger problem down the line.

I didn't have time to look into it much or find a possible solution to report this error to the user (as this should be considered a wrong usage of the jrl-cmakemodules).

Best

olivier-stasse:
Hi @gergondet 
Yes it is. But I think they are two problems.
The one you mention is indeed important and someone should probably look into that.
The other is simply that base.cmake is call before anything else and base.cmake calls itself shared-library,cmake. At this point there is no reason why CMAKE_LINKER is defined.
For some reasons not error was raised on my laptop even when removing the build directory. It only appeared on travis. 

Best,

bchretien:
Here (`--->` stands for "included by"):

`shared-library.cmake ---> pkg-config.cmake ---> base.cmake` 

And `base.cmake`  is supposed to be included first in the project's CMakeLists.txt. `CMakeFindBinUtils` however was not included anywhere in our code, and as @gergondet pointed out, that depends on the language set (included [there](https://github.com/Kitware/CMake/blob/3fe240a79a53d8069ab51c4a44c4e1f5062983f1/Modules/CMakeDetermineCXXCompiler.cmake#L172) for instance if C++ is chosen). At least now, no matter what happens, the CMake variable will be properly set.

bchretien:
As @gergondet pointed out, is there any particular reason for setting `PROJECT(...)` in the user code if we already do that in `base.cmake`?

gergondet:
Indeed, after investigating a bit, it seems that we overlooked something regarding the `PROJECT(...)`, according to the documentation https://cmake.org/cmake/help/v3.0/command/project.html:

> The top-level CMakeLists.txt file for a project must contain a literal, direct call to the project() command; > loading one through the include() command is not sufficient. If no such call exists CMake will implicitly > add one to the top that enables the default languages (C and CXX).

In our case, the call to `PROJECT(...)` in inside a macro inside an include and so it seems that it is not seen by cmake. As a result cmake chooses to enable the default languages (C and CXX) and does that before anything else in our CMakeLists file. In particular this sets `CMAKE_LINKER` (and actually this may be broken in older versions of cmake like the one coming in 12.04)

On the other having an explicit invokation of `PROJECT(...)` in the main cmake file triggers the language enabling at this point so if it's called after the inclusion of `cmake/base.cmake` you don't have `CMAKE_LINKER` and it fails.

While this PR fixes this particular issue, I think the better approach (other than letting people call `PROJECT(...)` before including `cmake/base.cmake` which I don't think is necessary) is to have `enable_language(CXX)` in `cmake/base.cmake`

olivier-stasse:
I suspect that the behavior is different for different cmake versions when PROJECT is not called inside the top level CMakeLists.txt. This will explain the different behaviors between12.04 LTS (cmake 2.8,7) and 14.04 LTS (cmake 3.3.2).
 
I like the suggestion from @gergondet if it really fixes the issue. I have been trying to follow the source code of cmake but I was not able to check if this is true. CMakeCXXCompiler.cmake.in is set CMAKE_LINKER but I do not completely master the connection between enable_language(CXX) and this file. If you have better information I would  :+1: this solution.
 
Anyway expecting that users will call PROJECT at the right place is not a good way to fix the issue.

BTW I want to thank you both for the nice work you are doing in maintaining jrl-cmakemodules and jrl-travis.

PR #75, opened 2016-01-27
Fix list of CPP stubs.

jmirabel:
I may have missed something, but I guess the variable `IDL_COMPILED_FILES` should be reset at each call.

I was getting a lot of warnings from `make` because of duplicated targets and some of the targets when not called.


bchretien:
:+1: Good catch!

PR #76, opened 2016-01-29
Properly handle boost components libs for pkg-config 

jcarpent:
Properly handle boost components libs for pkg-config and correct minor bug for handling OS X framework in pkg-config


bchretien:
It seems that this actually broke some stuff. The `PKG_CONFIG_APPEND_LIBS_RAW("${Boost_${UPPERCOMPONENT}_LIBRARY}")` should not be done there. One may want to look for Boost libraries, but not add them to the pkg-config file. For instance we now end up with `Libs: .... /usr/lib64/libboost_unit_test_framework.so` in our pkg-config files.

PR #78, opened 2016-03-06
boost.cmake: add new PKG_CONFIG_APPEND_BOOST_LIBS macro

bchretien:
This fixes the issue introduced in #76. Now Boost libraries can simply be added with:

```cmake
PKG_CONFIG_APPEND_BOOST_LIBS(filesystem system)
```

If there are specific treatments for OS X, they can be done in this macro.


gergondet:
Looks good to me. This is cleaner and simpler than current alternatives.

bchretien:
cc @jcarpent

jcarpent:
Looks also great for me. Thanks for this modification.

bchretien:
Tested with roboptim-core on Linux, everything looks fine.

jmirabel:
Shouldn't `PKG_CONFIG_APPEND_BOOST_LIBS(${BOOST_COMPONENTS})` be done automatically in SEARCH_FOR_BOOST macro ?

gergondet:
No. 

Typically, you need the test component if you are using Boost unit test framework but you shouldn't have it in your pkg-config file

PR #80, opened 2016-03-24
Update FINDPYTHON command to expose find_package options.

jmirabel:
This is useful to force a particular Python version when several are installed.


jmirabel:
Would it be a problem to remove the "hidden" `FINDPYTHON` in:
- `DYNAMIC_GRAPH_PYTHON_MODULE` https://github.com/jmirabel/jrl-cmakemodules/blob/master/python.cmake#L83
- `PYTHON_INSTALL_ON_SITE` https://github.com/jmirabel/jrl-cmakemodules/blob/master/python.cmake#L151
and for the user to call himself FINDPYTHON ?

This two "hidden" calls may lead to different python version be found.

A check for `PythonInterp_FOUND` could be done to print a useful message.

gergondet:
This looks reasonable but for backward compatibility and to avoid the issues you mentioned I think the argument should be optional (see [this](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/base.cmake#L151) for example)

Thanks

jmirabel:
Aren't they optional already ?
I'm not a cmake expert, I thought that with no arguments, `FIND_PACKAGE(PythonLibs ${ARGN})` is expanded to `FIND_PACKAGE(PythonLibs )`. That is not the case ?

gergondet:
My bad, this is exactly what is happening.

For the "hidden" `FINDPYTHON` I think you can make the check in `FINDPYTHON` itself to ensure that it's not called twice.

e.g. using `if(NOT DEFINED PYTHONINTERP_FOUND)`

jmirabel:
Sorry, I pushed a commit before seeing your answer. I updated to have the same behaviour as the one you mentioned.

haudren:
:+1: We already rely on this sort of behaviour but in a non standard way (See: https://github.com/jrl-umi3218/Tasks/blob/master/binding/python/CMakeLists.txt l.21). This change would allow us to make sure we get the correct version.

jmirabel:
> e/ As a remark, shouldn't the error messages now specify which python version has not been found ?
Indeed, it would make sense.

jmirabel:
There is still something a bit mysterious to me. According to [FindPythonLibs](https://cmake.org/cmake/help/v3.2/module/FindPythonLibs.html) and [FindPythonInterp](https://cmake.org/cmake/help/v3.2/module/FindPythonInterp.html) documentation:

> If also calling find_package(PythonLibs), call find_package(PythonInterp) first to get the currently active Python version by default with a consistent version of PYTHON_LIBRARIES.

This means:
- the calls `FIND_PACKAGE(PythonInterp)` and `FIND_PACKAGE(PythonLibs)` have to be inverted (which is not a problem)
- the following CMake code
```cmake
CMAKE_MINIMUM_REQUIRED(VERSION 2.8)

SET(Python_ADDITIONAL_VERSIONS 3.4)

FIND_PACKAGE(PythonInterp)
FIND_PACKAGE(PythonLibs)
```
gives the following output:
```
-- Found PythonInterp: /usr/bin/python3.4 (found version "3.4.3") 
-- Found PythonLibs: /usr/lib/i386-linux-gnu/libpython2.7.so (found version "2.7.6") 
```
which looks not good to me.
- the following CMake code
```cmake
CMAKE_MINIMUM_REQUIRED(VERSION 2.8)

INCLUDE(cmake/base.cmake)
INCLUDE(cmake/python.cmake)

FINDPYTHON(3.4 EXACT REQUIRED)
```
errors when not finding version 3.4 of PythonLibs (although it is properly installed on my computer) ...

Has someone any idea of the problem ?

jmirabel:
I solved part of the issue. I needed to install `libpython3.4-dev`. Sorry for that.

But I still think the second point of my last comment should return an error instead of just telling which two different versions it has found.

bchretien:
I've had this issue for a long time on Arch: the default Python is Python 3, ROS required Python 2, and CMake provided that kind of mismatching interpreter/libraries... To avoid this issue, we've had to force `PYTHON_EXECUTABLE`, `PYTHON_INCLUDE_DIR` and `PYTHON_LIBRARY` for a long time. However, I just checked again and it seems that it has finally been fixed. Apparently there has been a [lot of work done](https://github.com/Kitware/CMake/commits/master/Modules/FindPythonLibs.cmake) during the last year.

```cmake
SET(Python_ADDITIONAL_VERSIONS 2.7)
```
```
-- Found PythonInterp: /usr/bin/python2.7 (found version "2.7.11") 
-- Found PythonLibs: /usr/lib64/libpython2.7.so (found version "2.7.11") 
```

```cmake
SET(Python_ADDITIONAL_VERSIONS 3.3)
```
```
-- Found PythonInterp: /usr/bin/python (found version "3.5.1") 
-- Found PythonLibs: /usr/lib64/libpython3.5m.so (found version "3.5.1") 
```

Now I tested your changes to `python.cmake`, and it seems to be working:

```
2.7:

-- Found PythonLibs: /usr/lib64/libpython2.7.so (found suitable exact version "2.7.11") 
-- Found PythonInterp: /usr/bin/python2.7 (found suitable exact version "2.7.11") 

3.2 (without EXACT):

-- Found PythonLibs: /usr/lib64/libpython3.5m.so (found suitable version "3.5.1", minimum required is "3.2") 
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.5.1", minimum required is "3.2") 

3.4 (unavailable):

CMake Error at /usr/share/cmake-3.5/Modules/FindPackageHandleStandardArgs.cmake:148 (message):
  Could NOT find PythonLibs (missing: PYTHON_LIBRARIES PYTHON_INCLUDE_DIRS)
  (Required is exact version "3.4")
  /usr/share/cmake-3.5/Modules/FindPackageHandleStandardArgs.cmake:388 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.5/Modules/FindPythonLibs.cmake:265 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  cmake/python.cmake:29 (FIND_PACKAGE)
  CMakeLists.txt:10 (FINDPYTHON)

3.5:

-- Found PythonLibs: /usr/lib64/libpython3.5m.so (found suitable exact version "3.5.1") 
-- Found PythonInterp: /usr/bin/python3.5 (found suitable exact version "3.5.1")

Default:

-- Found PythonLibs: /usr/lib64/libpython3.5m.so (found version "3.5.1") 
-- Found PythonInterp: /usr/bin/python3.5 (found version "3.5.1")

**Comment by jmirabel** [comment]
Yes, I think it is ready to be merged now.

**Comment by jmirabel** [comment]
At the exception that I don't know how this behaves with CMake versions older than 3.2

**Comment by haudren** [comment]
I just tested with cmake 2.8.12.2 (Standard on Ubuntu 14.04):

- Calling with `EXACT` results in an error: 

> Could NOT find PythonInterp: Found unsuitable version "2.7.6", but required is exact version "2.7"

- I have python 3.4 and python3.4-dev installed, but cmake does not seem to be able to find python 3.4:

> Could NOT find PythonInterp: Found unsuitable version "2.7.6", but required is exact version "3.4"

- Calling directly `FIND_PACKAGE(PythonInterp 3.4)`, results, as predicted in an error as cmake can only find 2.7.6.

**Comment by jmirabel** [comment]
This is weird. I'm on Ubuntu 14.04 and my version of CMake is 3.2.2.
```bash
$ apt-cache show cmake
Package: cmake
Priority: optional
Section: devel
Installed-Size: 13762
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Architecture: i386
Version: 3.2.2-2ubuntu2~ubuntu14.04.1~ppa1
...

Package: cmake
Priority: optional
Section: devel
Installed-Size: 13296
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Modestas Vainius <modax@debian.org>
Architecture: i386
Version: 2.8.12.2-0ubuntu3
...
```

bchretien:
We can update the doc of the macro, and explain that for old versions of CMake, the underlying Python checks are known to be unreliable. Users can still increase the minimum required version of CMake as well if that's a critical feature.

Another option would be to provide our own recent copy of `FindPythonLibs.cmake`/`FindPythonInterp.cmake` (we already do this for `GNUInstallDirs.cmake`). It's clearly not ideal and may lead to other errors.

jmirabel:
I prefer the first option, with a warning in FINDPYTHON when `CMAKE_VERSION` is less than 3.2

jmirabel:
I figured out the CMake 3.2 version on my Ubuntu 14.04 comes from a repo specified by LAAS system administration, which is not by default on Ubuntu 14.04.
So the default version is 2.8.12 which means this may not work properly on Ubuntu 14.04...

> Another option would be to provide our own recent copy of `FindPythonLibs.cmake` / `FindPythonInterp.cmake`

That option may indeed be better.

haudren:
Actually, I think the best option is to check that the new behaviour is the same as the previous behaviour (I think we never really looked for anything different than 2.7). I think that rolling our own `FindPython{}.cmake` is *not* future-proof, and Ubuntu 16.04 will come out soon, and [ships](https://launchpad.net/ubuntu/xenial/+source/cmake) cmake 3.2. Thus a warning would probably be sufficient.

bchretien:
Hence the problem: people on Ubuntu will have issues when trying to build/install Python 3 projects, unless they update to the 16.04 or use a variant of the cmake package. Falling back on recent working copies of the scripts for old versions of CMake might not be worse than relying on macros that just aren't working at all. If we add a check on the CMake version, this would only impact old versions of CMake (e.g. pre 3.2), and more recent ones will still rely on their own `FindPython{}.cmake`.

haudren:
I hadn't seen that under that angle... Although it is unlikely that many python3 projects will be installed in 14.04 as of now, it may become prevalent once python3 becomes the default on 16.04 but the robot(s) will probably remain on 14.04. So, discard my previous message and let's ship our own `FindPython{}.cmake`!

jmirabel:
I agree with shipping the two files and using them only if CMake is older than 3.2.

I copied `FindPythonInterp.cmake` and `FindPythonLibs.cmake` from my computer (CMake version 3.2.2).
Can someone check this with an older version of CMake ?

haudren:
It seems that is still broken with my version of cmake (Finds the correct 3.4 libs, but only finds /usr/bin/python). I don't really have time to check more, I'll investigate tomorrow.

jmirabel:
I was able to test and fix with CMake 2.8.12. It now works with some points to be noted (see line notes).

python/FindPythonInterp.cmake
```diff
@@ -0,0 +1,180 @@
...
+    endif()
+    unset(_PYTHON_VERSION_RESULT)
+    unset(_VERSION)
+endif()
+
+# This is a hack to make FindPackageHandleStandardArgs not failing
+# when EXACT version of the form "3" or "3.2" is required and
+# PYTHONLIBS_VERSION_STRING is "3.2.4".
+# This hack is in CMake 3.2 and above.
+set(_PYTHON_VERSION_STRING "${PYTHON_VERSION_STRING}")
```
jmirabel: This hack should not cause any problem.

python/FindPythonInterp.cmake
```diff
@@ -0,0 +1,180 @@
...
+        set(PYTHON_VERSION_STRING "${PYTHON_VERSION_STRING}.${PythonInterp_FIND_VERSION_MINOR}")
+    endif(PythonInterp_FIND_VERSION_COUNT GREATER 1)
+    if(PythonInterp_FIND_VERSION_COUNT GREATER 2)
+        set(PYTHON_VERSION_STRING "${PYTHON_VERSION_STRING}.${PythonInterp_FIND_VERSION_TWEAK}")
+    endif(PythonInterp_FIND_VERSION_COUNT GREATER 2)
+endif()
+
+# handle the QUIETLY and REQUIRED arguments and set PYTHONINTERP_FOUND to TRUE if
+# all listed variables are TRUE
+include(FindPackageHandleStandardArgs)
```
jmirabel: This line was originally
```cmake
include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)
```
I'm not sure of why they including it this way. I guess it was to make sure `include` finds the good version.

python/FindPythonInterp.cmake
bchretien:
```cmake
# This hack is in CMake 3.2 and above.
```

Is it really in CMake? I couldn't find anything like this in CMake's current master branch.

jmirabel: Yes, in this commit:
https://github.com/Kitware/CMake/commit/4f9bf4468bd5bcd48be080fa0bc736ebb9df4769

The change is in `FindPackageHandleStandardArgs.cmake`

bchretien: Ah, I didn't realize this was from another module. Thanks for clearing that up!

haudren:
Ok, I tested this on my computer with cmake 2.8: cmake now finds the correct combination of library and interpreter. However, some weird caching is afoot:
> FINDPYTHON(2.7 EXACT REQUIRED)
> -- Found PythonInterp: /usr/bin/python2.7 (found suitable exact version "2.7") 
> -- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython2.7.so (found suitable exact version "2.7") 
> Modify CMakeLists.txt
> FINDPYTHON(3.4 EXACT REQUIRED)
> -- Found PythonInterp: /usr/bin/python2.7 (found suitable exact version "3.4") 
> -- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython2.7.so (found suitable exact version "3.4") 
> Remove CMakeCache
> FINDPYTHON(3.4 EXACT REQUIRED)
> -- Found PythonInterp: /usr/bin/python3.4 (found suitable exact version "3.4") 
> -- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython3.4m.so (found suitable exact version "3.4") 
I don't think that this is a big issue anyway, but it is a little bit strange, as always with CMakeCache :(

jmirabel:
I had the same issue. I didn't try to solve it. There might be a bug in the FindPythonLibs and/or FindPythonInterp files but I'm not sure it's worth digging it up...

haudren:
No I think it's all right, this is ready to be merged in my opinion :)

bchretien:
Looks ok to me ;)

jcarpent:
:+1:

PR #81, opened 2016-03-25
idl.cmake: add missing omniidl flag

bchretien:
This solves the following error when using the generated header/library:

``` sh
error: no match for ‘operator<<=’ (operand types are ‘CORBA::Any’ and ‘...’)
```

Followed by thousands of `no known conversion for argument 1 from ‘...’ to ‘cdrStream&’`. This flag is actually used in `idlrtc.cmake`, but in practice we all seem to be using `idl.cmake`'s macros. Note that the documentation says:

``` txt
C++ BACK-END OPTIONS
        ...
       -Wba           Generate definitions for TypeCode and Any.
```


haudren:
Looks nice ! Though, it brings the question: should we get rid of idlrtc ?

Compared to idlrtc, we also dropped `-Wbuse_quotes    Use quotes in #includes: "foo" rather than <foo>`, but I don't think this is necessary for now. Moreover, with your change, it should be easy to specify extra flags, somtehing along the lines of:
```cmake
  SET(IDL_FLAGS "-Wbkeep_inc_path" "-Wba" ${IDL_EXTRA_FLAGS})
```

bchretien:
We can certainly allow extra flags, although the user could break the code if non-default suffixes are used for the generated files (e.g. `SET(IDL_EXTRA_FLAGS "-Wbh=Foo.hh")`).

As for `idlrtc.cmake`, we can still keep the macro for backwards compatibility, and replace its implementation with a call to `GENERATE_IDL_CPP` with adequate extra parameters (if the differences are indeed required).

haudren:
Good point... I think it could be nice to merge the two of them, but we need to make sure the user does not break anything, How about checking for one of the options that modify the suffixes ?
```cmake
if(${IDL_EXTRA_FLAGS} MATCHES -Wb[hsd])
    MESSAGE(You hsould not modify the deffault suffixes)
endif()
```
Note that we could even a two layer implementation:
- A user-friendly level that rejects the suffix-modifying options, and uses the defaults
- A lower-level macro that takes as arguments the suffixes

jcarpent:
#180 seems to solve this PR.

PR #82, opened 2016-04-11
Include optional version argument and "support" for Windows

aescande:


gergondet:
Looks good, thanks for the contribution.

However, I'd rather have the manual search as a fallback in case pkg-config fails to find eigen on Windows (or there is no pkg-config installed on the Windows installation) .

I rely on pkg-config to find eigen on Windows platform by setting up the .pc my self ([e.g.](https://github.com/jrl-umi3218/jrl-travis/blob/master/dependencies/eigen-3.2.ps1))

The manual search should also handle the case when eigen is not found.

bchretien:
:+1: LGTM (pkg-config + findpath for all the OSes).

eigen.cmake
```diff
@@ -17,24 +17,34 @@
...
+  PKG_CHECK_MODULES(_Eigen ${Eigen_REQUIRED})
   
-  SET(${PROJECT_NAME}_CXX_FLAGS "${${PROJECT_NAME}_CXX_FLAGS} ${_Eigen_CFLAGS}")
-  SET(${PROJECT_NAME}_LINK_FLAGS "${${PROJECT_NAME}_LINK_FLAGS} ${_Eigen_LDFLAGS}")
-
-  STRING(REGEX REPLACE "-I" "" Eigen_INCLUDE_DIR ${_Eigen_CFLAGS})
-  INCLUDE_DIRECTORIES(SYSTEM ${Eigen_INCLUDE_DIR} )
+  IF(${_Eigen_FOUND})
+    SET(${PROJECT_NAME}_CXX_FLAGS "${${PROJECT_NAME}_CXX_FLAGS} ${_Eigen_CFLAGS}")
+    SET(${PROJECT_NAME}_LINK_FLAGS "${${PROJECT_NAME}_LINK_FLAGS} ${_Eigen_LDFLAGS}")
```
bchretien: Quick note: we should probably use `_Eigen_CFLAGS_OTHER` and `_Eigen_LDFLAGS_OTHER` here.

eigen.cmake
```diff
@@ -17,24 +17,34 @@
...
+    
+    STRING(REGEX REPLACE "-I" "" Eigen_INCLUDE_DIR ${_Eigen_CFLAGS})
+    INCLUDE_DIRECTORIES(SYSTEM ${Eigen_INCLUDE_DIR} )
```
bchretien: Can't we use `_Eigen_INCLUDE_DIRS` directly here?

eigen.cmake
```diff
@@ -17,24 +17,34 @@
 # -----------------
 #
 # This macro gets eigen include path from pkg-config file, and adds it include directories.
+# On Windows, there is no pkg-config for Eigen, thus we fall back on a manual search.
```
gergondet: This is out of date given the recent change
 Please update it to reflect the current strategy

gergondet:
:+1:

PR #83, opened 2016-04-12
pkg-config.cmake: fix for duplicated CFLAGS

bchretien:
We noticed a few things:

* ~~`SET_TARGET_PROPERTIES(${TARGET} PROPERTIES COMPILE_OPTIONS "${CFLAGS}")` **appends** the flags, it does not replace them (cf. the doc for [2.8.12](https://cmake.org/cmake/help/v2.8.12/cmake.html#command:set_target_properties) or this [SO question](http://stackoverflow.com/questions/5096881/does-set-target-properties-in-cmake-override-cmake-cxx-flags)). Thus, we don't need to get them, since that would lead to duplicated flags.~~
* The `INCLUDE_DIRS` variable contains include directories obtained by parsing the `CFLAGS`, cf. the CMake code in `FindPkgConfig.cmake`:

```cmake
_pkgconfig_invoke_dyn("${_pkg_check_modules_packages}" "${_prefix}" INCLUDE_DIRS        "(^| )-I" --cflags-only-I )
_pkgconfig_invoke_dyn("${_pkg_check_modules_packages}" "${_prefix}" CFLAGS              ""        --cflags )
_pkgconfig_invoke_dyn("${_pkg_check_modules_packages}" "${_prefix}" CFLAGS_OTHER        ""        --cflags-only-other )
```

To avoid having duplicated include directories (added by both CMake's `INCLUDE_DIRECTORIES(${FOO_INCLUDE_DIRS})` and the `CFLAGS`), we need to use `CFLAGS_OTHER` instead (basically, `CFLAGS = INCLUDE_DIRS  + CFLAGS_OTHER`). Note that if the `.pc` contains an entry in the `CFLAGS` with `-isystem`, it will keep on being treated as a `CFLAGS` entry, since they check explicitly for `-I`.


bchretien:
Ok apparently, the following sentence:

> The LINK_FLAGS property can be used **to add extra flags** to the link step of a target. LINK_FLAGS_\<CONFIG\> will **add** to the configuration \<CONFIG\>, for example, DEBUG, RELEASE, MINSIZEREL, RELWITHDEBINFO. [...] The COMPILE_FLAGS property sets additional compiler flags used to build sources within the target. It may also be used to pass additional preprocessor definitions.

does not mean appending, so CFLAGS and LDFLAGS are not treated in the same way... I'll update the PR and make some more tests.

bchretien:
Ok, that's my bad. There are several properties (`COMPILE_FLAGS`, `COMPILE_OPTIONS` etc.), they are all appended to the generated compilation command, but need to be copied if we want to update them, so that part of the `pkg-config.cmake` was right, and both the documentation and the SO answers did not tackle that issue. We could however use `target_compile_options` which provides more options, and that ones does append:

> Specify compile options to use when compiling a given target. The named <target> must have been created by a command such as add_executable() or add_library() and must not be an IMPORTED Target. If BEFORE is specified, the content will be prepended to the property instead of being appended.

but it is only available for CMake >= 2.8.12. So I removed the commit related to that, and the switch to `*FLAGS_OTHER` is still relevant.

bchretien:
We made some more tests, including on Windows, and it actually solved issues we had with the include directories: another parsing of the `CFLAGS` to find include directories appears to be done, probably in the Visual Studio generator, and it prepended them to the list of include directories. Thus system include directories were used before local include directories.

So now, the whole include directory support is cleaner, but we can probably improve things a bit more, e.g. by adding a sanity check to `SETUP_PROJECT_FINALIZE` to make sure that `CFLAGS` do not contain include directories (`CMAKE_CXX_FLAGS`, per-target `COMPILE_OPTIONS/COMPILE_DEFINITIONS`, etc.), and warn the user if that's the case.

bchretien:
Apparently there is no non-intrusive way to loop over all the targets to make some sanity checks, so we can forget about that. PR ready for merge.

haudren:
:+1:

PR #86, opened 2016-04-22
Adding MSVC specific macro

aescande:


msvc-specific.cmake
```diff
@@ -0,0 +1,122 @@
...
+# or when no MSVC compiler or an unknown version of it is used.
+#
+# A reason for the version to be unknown is that it is newer than the versions
+# recognize by the above macro GET_MSVC_VERSION. In this case, please update 
+# the macro and its documentation.
+#
+MACRO(REQUEST_MINIMUM_MSVC_VERSION)
+  if(NOT ${ARGC} GREATER 0)
+    message("REQUEST_MINIMUM_MSVC_VERSION requires an argument" FATAL_ERROR)
+  endif()
```
gergondet: CMake macros accept arguments so you could write
`MACRO(REQUEST_MINIMUM_MSVC_VERSION VERSION)`
and forgo the argument checking you're doing in this macro and the next

msvc-specific.cmake
```diff
@@ -0,0 +1,122 @@
...
+# ---------------------------------
+#
+# Return a fatal error if the current MSVC version is strictly lower than VER
+# or when no MSVC compiler or an unknown version of it is used.
+#
+# A reason for the version to be unknown is that it is newer than the versions
+# recognize by the above macro GET_MSVC_VERSION. In this case, please update 
+# the macro and its documentation.
+#
+MACRO(REQUEST_MINIMUM_MSVC_VERSION)
```
gergondet: `REQUIRE_MINIMUM_MSVC_VERSION` would better imho

msvc-specific.cmake
```diff
@@ -0,0 +1,122 @@
...
+  if(NOT ${ARGC} GREATER 0)
+    message("REQUEST_MINIMUM_MSVC_VERSION requires an argument" FATAL_ERROR)
+  endif()
+  GET_MSVC_VERSION()
+  if (${MSVC_VERSION})
+    if (${MSVC_VERSION} STRGREATER ${ARGV0})
+    else()
+	  message("Minimum MSVC version required is " ${ARGV0} FATAL_ERROR)
```
gergondet: Maybe it's personal but I find the numbering of Visual Studio version to be confusing (e.g. 12.0 is Visual Studio 2013, 14.0 is Visual Studio 2015 and 15.0 is Visual Studio 15) so I think it would be very helpful to report the version that was detected along with this failure.

msvc-specific.cmake
```diff
@@ -0,0 +1,122 @@
...
+# recognize by the above macro GET_MSVC_VERSION. In this case, please update 
+# the macro and its documentation.
+#
+MACRO(REQUEST_MINIMUM_MSVC_VERSION)
+  if(NOT ${ARGC} GREATER 0)
+    message("REQUEST_MINIMUM_MSVC_VERSION requires an argument" FATAL_ERROR)
+  endif()
+  GET_MSVC_VERSION()
+  if (${MSVC_VERSION})
+    if (${MSVC_VERSION} STRGREATER ${ARGV0})
```
gergondet: why not
`if (NOT ${MSVC_VERSION} STRGREATER ${ARGV0})`
?

gergondet: Also, `VERSION_GREATER` is the correct comparison here (with `STRGREATER` you get MSVC6 > MSVC14)

msvc-specific.cmake
```diff
@@ -0,0 +1,122 @@
...
+    set(MSVC_VERSION "10.0")
+  elseif(MSVC11)
+    set(MSVC_VERSION "11.0")
+  elseif(MSVC12)
+    set(MSVC_VERSION "12.0")
+  elseif(MSVC14)
+    set(MSVC_VERSION "14.0")
+  else()
+    set(MSVC_VERSION "")
+	message("MSVC version not found: you're either targeting a non msvc compiler or you're using a new compiler, in which case, please update the GET_MSVC_VERSION macro")
```
gergondet: I don't think this approach is good given the context in which jrl-cmakemodules are used (i.e. as a git submodule)

I think `MSVC` is always set by cmake whatever the version is used so if `MSVC` is set but none of the other ones (i.e. versions known by this macro) then we can set `MSVC_VERSION` to 14.0 or even 14.5. It's very unlikely at this point that cmake will start providing generators for MSVC < 6.0

msvc-specific.cmake
aescande: The line is comparing strings like "12.0" and "7.1", so I think this is correct here

msvc-specific.cmake
```diff
@@ -0,0 +1,122 @@
...
+      file(APPEND ${DOT_USER_FILE_PATH} "  " <PropertyGroup " " Condition=\"'\$(Configuration) | \$(Platform) ' =='${config}|${MSVC_COMPILER_VERSION}'\">\n)
+	  file(APPEND ${DOT_USER_FILE_PATH} "    " <LocalDebuggerEnvironment>PATH=\$(SolutionDir) \\ src \\ \$ (Configuration) \; %PATH%</LocalDebuggerEnvironment>\n)
+	  file(APPEND ${DOT_USER_FILE_PATH} "    " <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>\n)
+	  file(APPEND ${DOT_USER_FILE_PATH} "  " </PropertyGroup>\n)
+    ENDFOREACH()
+	file(APPEND ${DOT_USER_FILE_PATH} </Project>)
+  else(MSVC)
+    message("This macro is intended for MSVC only, skipping .user file generation")
+  endif(MSVC)
+ENDMACRO(GENERATE_DOT_USER_FILE)
```
gergondet: This is great but you could use configure_file to generate the .user file here (several modules in jrl-cmakemodules make the assumption that the module is located under the `cmake` folder for this)

msvc-specific.cmake
aescande: As far as I could find in the documentation, CMake is not providing generators for MSVC<6.0, so your trick would work.
However, I think it is better to force someone using a new generator to update this macro. Otherwise, in a few years, we might encounter problems where REQUIRE_MINIMUM_MSVC_VERSION("16.0") would accept a 15.0 generator

msvc-specific.cmake
gergondet:
```
if("7.1" STRGREATER "12.0")
  MESSAGE("OK")
endif()
```

Prints OK but `VERSION_GREATER` does what is expected

aescande: OK, I misunderstood STRGREATER

gergondet:
Looks good to me 👍

PR #87, opened 2016-04-27
eigen.cmake: fix Eigen dependency in generated .pc

bchretien:
As @francois-keith pointed out [here](https://github.com/roboptim/roboptim-core/pull/110#discussion_r61175686), using `SEARCH_FOR_EIGEN` does not add Eigen as a required dependency in the generated `${PROJECT_NAME}.pc` if the pkg-config lookup was used, and it does not add the include directory to the `CFLAGS` if the "manual" search was used. This PR is an attempt to solve both these issues.

@aescande could you test this on Windows?

### DIFF
--- a/eigen.cmake
+++ b/eigen.cmake
@@ -29,15 +29,16 @@ MACRO(SEARCH_FOR_EIGEN)
   ELSEIF(NOT Eigen_REQUIRED)
     SET(Eigen_REQUIRED "eigen3 >= 3.0.0")
   ENDIF()
-  
+
   # looking for .pc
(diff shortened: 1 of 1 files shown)


haudren:
:+1: I actually encountered the bug in some situations with mc-rtc, but did not investigate! Thanks

aescande:
With this, I have eigen appearing in the Cflags of roboptim.pc

haudren:
I just tested this with SpaceVecAlg: it does add Eigen >= 3.2.0 to the requirements of the .pc, and it does not add the Eigen include dir to the CFLAGS !

Everything looks good :cake:

gergondet:
Good job :+1:

PR #92, opened 2016-06-14
[python] Update cmake policy

olivier-stasse:
Wrap target is not allowed according to cmake policy CMP0037.
But this should not clash with intermediate cmake targets unless
there is a major python update in generating bindings.
Therefore this commit set this policy to OLD when generating python plugin.


olivier-stasse:
Commit c7c0874 add cmake_policy(push) and cmake_policy(pop) to guarantee that the cmake_policy is kept. From a suggestion by @gergondet .

PR #94, opened 2016-06-24
Handle linking against boost_python on APPLE

jcarpent:
On Apple system, boost_python has unresolved symbols and are resolved according to the Python interpreter version.
One has to use `TARGET_LINK_BOOST_PYTHON ` to properly links its application against boost_python.

### DIFF
--- a/boost.cmake
+++ b/boost.cmake
@@ -82,7 +82,38 @@ MACRO(SEARCH_FOR_BOOST)
       Boost_${UPPERCOMPONENT}_LIBRARY_RELEASE
       )
   ENDFOREACH()
+
+  # On darwin systems, we must link againt boost_python with unresolved symbols.
+  # We then remove boost_python from the global Boost_LIBRARIES list to handle it with specific care.
+  IF(${Boost_PYTHON_LIBRARY})
+    LIST(REMOVE_ITEM Boost_LIBRARIES ${Boost_PYTHON_LIBRARY})
+  ENDIF(${Boost_PYTHON_LIBRARY})
+
 ENDMACRO(SEARCH_FOR_BOOST)
+
+# TARGET_LINK_BOOST_PYTHON
+# ------------------------
+#
+# Link target againt boost_python library.
+# \input target is either a library or an executable
+# On darwin systems, boost_python is not linked against any python library.
+# This linkage is resolved at execution time via the python interpreter.
+# We then need to stipulate that boost_python has unresolved symbols at compile time.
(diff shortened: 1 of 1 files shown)


boost.cmake
```diff
@@ -99,8 +128,10 @@ MACRO(PKG_CONFIG_APPEND_BOOST_LIBS)
 
   FOREACH(COMPONENT ${ARGN})
     IF(APPLE)
-      STRING(TOUPPER ${COMPONENT} UPPERCOMPONENT)
-      PKG_CONFIG_APPEND_LIBS_RAW(${Boost_${UPPERCOMPONENT}_LIBRARY})
+      #STRING(TOUPPER ${COMPONENT} UPPERCOMPONENT)
+      #PKG_CONFIG_APPEND_LIBS_RAW(${Boost_${UPPERCOMPONENT}_LIBRARY})
```
bchretien: Dead code.

boost.cmake
```diff
@@ -82,8 +82,37 @@ MACRO(SEARCH_FOR_BOOST)
       Boost_${UPPERCOMPONENT}_LIBRARY_RELEASE
       )
   ENDFOREACH()
+
+  # On darwin systems, we must link againt boost_python with unresolved symbols.
+  # We then remove boost_python from the global Boost_LIBRARIES list to handle it with specific care.
+  LIST(REMOVE_ITEM Boost_LIBRARIES ${Boost_PYTHON_LIBRARY})
```
bchretien: Could that break things on non-Apple systems?

boost.cmake
jcarpent: You're totally right.

boost.cmake
jcarpent: I do not think so. The only thing to do is to use macro `TARGET_LINK_BOOST_PYTHON` to link against boost_python. It will automatically properly links again boost_python according to the OS (https://github.com/boostorg/build/pull/78).

bchretien: I guess there's no way around this anyway. jrl-cmakemodules users with code linking with Boost Python should be careful when updating (cc @gergondet / @haudren, although AFAIK nobody's concerned in our team).

jcarpent:
I also suggest to print a warning message to insist on the new strategy.

jcarpent:
@bchretien I think this PR is ready to be merged to the master branch. Do you see any disagreement?

bchretien:
:+1: LGTM!

PR #97, opened 2016-07-24
[Win32] Correct pc file when compiling with Visual.

francois-keith:
More specifically, the CMAKE_CONFIGURATION_TYPE variable is always defined when generating the project for Visual Studio (it equals Release,RelWithDebInfo,Debug...).
With the former version of the code, if one run cmake with the option CMAKE_BUILD_TYPE defined (to Release or Debug), the library specified in the .pc file would always link with the debug version of the lib.


gergondet:
This change would break many VS projects. `CMAKE_CONFIGURATION_TYPES` is defined for generators that produce multiple build configurations (such as the Visual Studio one). You shouldn't define `CMAKE_BUILD_TYPE` when you use such a generator as the build type is decided through the interface (and perhaps the CLI?). If you set something based solely on the provided value of `CMAKE_BUILD_TYPE` then the project is potentially broken for every other configuration.

If you still want to define `CMAKE_BUILD_TYPE` with such generators and if you want to fix the .pc generation then the correct fix would be to check whether `CMAKE_BUILD_TYPE` is defined/non-empty and give that a priority over `CMAKE_CONFIGURATION_TYPES` (but as I said above, the project would be in a broken state).

francois-keith:
I see.
I did this to select the .pc file installed, but I didn't notice that both (.pc and _d.pc) were installed using the MSVC INSTALL target. My mistake, I cancel this PR.

PR #98, opened 2016-07-26
Correct typo.

francois-keith:


gergondet:
Nice catch, thanks :smile_cat:

Issue #66, opened 2015-09-24
Update Wiki and/or provide useful readme


bchretien:
A list of the main features (e.g. Doxygen support) would be really helpful. Right now, people need to read the `*.cmake` files. This could be made available in the Wiki, or in the `README.md` (which would probably be smarter since this is bundled in the submodule).

bchretien:
Another possibility would be to generate the documentation with sphinx, directly from the `*.cmake` files. This seems to be the approach followed by CMake itself (i.e. [this](https://github.com/Kitware/CMake/blob/acd8a73044e879286f0cfa24b54497a8307f204b/Modules/FindCUDA.cmake#L1-L323) leads to [that](https://cmake.org/cmake/help/v3.5/module/FindCUDA.html?highlight=findcuda)).

jmirabel:
I tried the option mentioned above. Using `sphinx` and the `cmake.py` extension from the [cmake repo](https://gitlab.kitware.com/cmake/cmake/blob/master/Utilities/Sphinx/cmake.py), I am able to generate a doc from the cmake file by adding line:
```
#.rst:
```
before the documentation section. The documentation looks like on the link above.

If we go for this solution, the question is: where could the documentation be stored ? on a branch `gh-pages` on this repo, which a github hook to update it on each merge ?

jmirabel:
As an example, generated from my fork, it generates this:
- http://github.com/jmirabel/jrl-cmakemodules
- http://cmake-submodule.readthedocs.io/en/latest/

Note that:
- the files that are parsed must be specified manually (see `docs/pages/external_deps.rst` for instance)
- only the macros where I added `#.rst:` are documented

jmirabel:
> I think the documentation headline (e.g. SEARCH_FOR_EIGEN here) could include the function/macro "prototype"

Feel free to edit.

> Btw, what is the effect of the command/variable precision in the documentation? I don't see any effect on the display of the doc. I also didn't see any option documented in the example you provided. Is this is a limitation of cmake.py?

From my understanding of `cmake.py`, the only difference are:
- references to commands are case insensitive while references to variables are case sensitive;
- they appear differently in the index;
- in the HTML, they have different classes so it should not be hard customize their appearance.

> I'm also wondering if it is possible to split internals/public documentation if they originate from the same file (e.g. pkg-config.cmake has both internal and public macros that are worth documenting).

[...]

PR #99, opened 2016-09-29
Make doc optional if INSTALL_DOCUMENTATION is set

jcarpent:


gergondet:
This looks good to me.

Also a reminder (to myself ^^) concerning #66 as this kind of option should be documented.

jcarpent:
Thanks !

PR #102, opened 2016-10-13
Add macro ADD_SEPARATE_DEPENDENCY

jcarpent:
It allows the search of requiered depencies which will not be added to the list of depencies
of the current project.


jmirabel:
I would rather call it `ADD_COMPILE_DEPENDENCY` because you want is a dependency required for compiling but not for linking to your library.

jcarpent:
Thank you for your feedback. I just integrated your suggestion.

gergondet:
Looks good to me, this is pretty much a rename of `ADD_DOC_DEPENDENCY`with a better name.

Good job :+1:

Issue #91, opened 2016-06-01
Version number is not computed if git is not found


aescande:
This is half a problem of jrl-cmakemodules, half a problem of the way we write our CMakeLists.

As of now, the version number of a library (that will be used to generate the .pc) can be computed solely based on git tags. This gives a rich version number which includes the exact commit used for generating the library.
Because we rely on this behavior, we (almost?) never set PROJECT_VERSION in the CMakeLists.

As a consequence, if git is not found (and there are several user scenarios in which this can happen), the version will be set to UNKNOWN-dirty. This is problematic for subsequent use of a library, when a minimum version number is required.

One solution would be to simply set PROJECT_VERSION in the library's CMakeLists, but, due to the actual logic in base.cmake and version.cmake, it will override the richer git-computed version number, changing the actual number whether git is found or not.
The alternative is to change this logic to have the git-computed version take precedence over PROJECT_VERSION, and fallback to PROJECT_VERSION only when git is not found. Possibly, we can add an error when neither git nor PROJECT_VERSION is found to enforce the generation of a number.

If nobody objects, I'll make the changes of the alternative.

olivier-stasse:
:+1:

thomas-moulard:
disclaimer: giving out some context and my personal opinion, I am no involved in this project anymore so please feel free to do whatever seems best for you :)

version.cmake line 58 to 61 look for a .version at the root of the project which has the precedence over the git information.

It may be broken but the major problem with using PROJECT_VERSION is that if you use the current scheme then the version number depends on the number of commit since the last tag. And if PROJECT_VERSION is set into the CMakeLists.txt then you just cannot have coherent versioning as you need to commit this information which will in turn change the version number.

The approach I would recommend is instead to just generate this file when needed.

The current approach to generate this file is here:
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/dist.cmake#L44

It is Linux-specific but making a similar system for Windows should not be too hard.

Also, the initial reason why the PROJECT_VERSION has not been chosen is that I am quite sure that no one will maintain this flag, so as no information is better than bad information, I would rather go with setting 0.0 as the default PROJECT_VERSION in case the version cannot be determined.

bchretien:
> As a consequence, if git is not found (and there are several user scenarios in which this can happen), the version will be set to UNKNOWN-dirty.

Could you clarify these scenarios?

We can distinguish 2 cases of installed packages:
* release version,
* development version.

[...]

PR #103, opened 2016-10-26
Can now generate standard CMake *Config.cmake and *Version.cmake files.

ahundt:
based on: https://github.com/forexample/package-example

This helps generate config files so other regular cmake packages can find and use library dependencies. For example a simple change to [RBDyn/src/CMakeLists.txt](https://github.com/jrl-umi3218/RBDyn/blob/master/src/CMakeLists.txt) will install the following:

```
-- /usr/local/lib/cmake/RBDyn/RBDynConfig.cmake
-- /usr/local/lib/cmake/RBDyn/RBDynConfigVersion.cmake
-- /usr/local/lib/cmake/RBDyn/RBDynTargets.cmake
-- /usr/local/lib/cmake/RBDyn/RBDynTargets-noconfig.cmake
```

That way a call to find_package(RBDyn) will automatically be configured!

Here is an example change to [RBDyn/src/CMakeLists.txt](https://github.com/jrl-umi3218/RBDyn/blob/master/src/CMakeLists.txt), which I will submit separately once this is merged:

``` cmake
set(SOURCES MultiBodyGraph.cpp MultiBody.cpp MultiBodyConfig.cpp
  FK.cpp FV.cpp FA.cpp Jacobian.cpp ID.cpp IK.cpp IS.cpp FD.cpp EulerIntegration.cpp
  CoM.cpp Momentum.cpp ZMP.cpp IDIM.cpp VisServo.cpp)
set(HEADERS Body.h Joint.h MultiBodyGraph.h MultiBody.h MultiBodyConfig.h
  FK.h FV.h FA.h Jacobian.h ID.h IK.h IS.h FD.h EulerIntegration.h CoM.h
  Momentum.h ZMP.h IDIM.h VisServo.h)

add_library(RBDyn SHARED ${SOURCES} ${HEADERS})
PKG_CONFIG_USE_DEPENDENCY(RBDyn SpaceVecAlg)
set_target_properties(RBDyn PROPERTIES COMPILE_FLAGS "-Drbdyn_EXPORTS")
set_target_properties(RBDyn PROPERTIES SOVERSION 0 VERSION 0.9.0)


# Targets:
#   * <prefix>/lib/libbar.a
#   * <prefix>/lib/libbaz.a
#   * header location after install: <prefix>/include/foo/Bar.hpp
#   * headers can be included by C++ code `#include <foo/Bar.hpp>`
install(
    TARGETS RBDyn
    EXPORT "${targets_export_name}"
    LIBRARY DESTINATION "lib"
    ARCHIVE DESTINATION "lib"
    RUNTIME DESTINATION "bin"
    INCLUDES DESTINATION "${include_install_dir}"
)
install(FILES ${HEADERS} DESTINATION ${include_install_destination})
```

### DIFF
--- a/base.cmake
+++ b/base.cmake
@@ -121,16 +121,16 @@ MACRO(_SETUP_PROJECT_PACKAGE_INIT)
 #   * <prefix>/lib/cmake/<PROJECT-NAME>
 #   * <prefix>/lib/
 #   * <prefix>/include/
-set(config_install_dir "lib/cmake/${PROJECT_NAME}")
-set(include_install_dir "include")
-set(include_install_destination "${include_install_dir}/${PROJECT_NAME}")
+set(CONFIG_INSTALL_DIR "lib/cmake/${PROJECT_NAME}")
+set(INCLUDE_INSTALL_DIR "include")
+set(INCLUDE_INSTALL_DESTINATION "${INCLUDE_INSTALL_DIR}/${PROJECT_NAME}")
 
-set(generated_dir "${CMAKE_CURRENT_BINARY_DIR}/generated")
+set(GENERATED_DIR "${CMAKE_CURRENT_BINARY_DIR}/generated")
 
 # Configuration
-set(version_config "${generated_dir}/${PROJECT_NAME}ConfigVersion.cmake")
-set(project_config "${generated_dir}/${PROJECT_NAME}Config.cmake")
-set(targets_export_name "${PROJECT_NAME}Targets")
+set(VERSION_CONFIG "${GENERATED_DIR}/${PROJECT_NAME}ConfigVersion.cmake")
+set(PROJECT_CONFIG "${GENERATED_DIR}/${PROJECT_NAME}Config.cmake")
+set(TARGETS_EXPORT_NAME "${PROJECT_NAME}Targets")
 set(namespace "${PROJECT_NAME}::")
 ENDMACRO(_SETUP_PROJECT_PACKAGE_INIT)
 
@@ -232,16 +232,16 @@ MACRO(_SETUP_PROJECT_PACKAGE_FINALIZE)
 #   * <prefix>/lib/cmake/<PROJECT-NAME>
 #   * <prefix>/lib/
 #   * <prefix>/include/
-set(config_install_dir "lib/cmake/${PROJECT_NAME}")
-set(include_install_dir "include")
-set(include_install_destination "${include_install_dir}/${PROJECT_NAME}")
+set(CONFIG_INSTALL_DIR "lib/cmake/${PROJECT_NAME}")
+set(INCLUDE_INSTALL_DIR "include")
+set(INCLUDE_INSTALL_DESTINATION "${INCLUDE_INSTALL_DIR}/${PROJECT_NAME}")
 
-set(generated_dir "${CMAKE_CURRENT_BINARY_DIR}/generated")
+set(GENERATED_DIR "${CMAKE_CURRENT_BINARY_DIR}/generated")
 
 # Configuration
-set(version_config "${generated_dir}/${PROJECT_NAME}ConfigVersion.cmake")
-set(project_config "${generated_dir}/${PROJECT_NAME}Config.cmake")
-set(targets_export_name "${PROJECT_NAME}Targets")
+set(VERSION_CONFIG "${GENERATED_DIR}/${PROJECT_NAME}ConfigVersion.cmake")
+set(PROJECT_CONFIG "${GENERATED_DIR}/${PROJECT_NAME}Config.cmake")
+set(TARGETS_EXPORT_NAME "${PROJECT_NAME}Targets")
 set(namespace "${PROJECT_NAME}::")
 
 # Include module with fuction 'write_basic_package_version_file'
@@ -250,66 +250,33 @@ include(CMakePackageConfigHelpers)
 # Configure '<PROJECT-NAME>ConfigVersion.cmake'
 # Note: PROJECT_VERSION is used as a VERSION
 write_basic_package_version_file(
-    "${version_config}" COMPATIBILITY SameMajorVersion
+    "${VERSION_CONFIG}" COMPATIBILITY SameMajorVersion
 )
 
 # Configure '<PROJECT-NAME>Config.cmake'
 # Use variables:
-#   * targets_export_name
+#   * TARGETS_EXPORT_NAME
 #   * PROJECT_NAME
 configure_package_config_file(
     "cmake/Config.cmake.in"
-    "${project_config}"
-    INSTALL_DESTINATION "${config_install_dir}"
+    "${PROJECT_CONFIG}"
+    INSTALL_DESTINATION "${CONFIG_INSTALL_DIR}"
 )
-
-# # Targets:
-# #   * <prefix>/lib/libbar.a
-# #   * <prefix>/lib/libbaz.a
-# #   * header location after install: <prefix>/include/foo/Bar.hpp
-# #   * headers can be included by C++ code `#include <foo/Bar.hpp>`
-# install(
-#     TARGETS ${SETUP_PROJECT_PACKAGE_TARGETS}
-#     EXPORT "${targets_export_name}"
-#     LIBRARY DESTINATION "lib"
-#     ARCHIVE DESTINATION "lib"
-#     RUNTIME DESTINATION "bin"
-#     INCLUDES DESTINATION "${include_install_dir}"
-# )
-
-# # Headers:
-# #   * Source/foo/Bar.hpp -> <prefix>/include/foo/Bar.hpp
-# #   * Source/foo/Baz.hpp -> <prefix>/include/foo/Baz.hpp
-# install(
-#     DIRECTORY "Source/foo"
-#     DESTINATION "${include_install_dir}"
-#     FILES_MATCHING PATTERN "*.h*"
-# )
-
-# # Export headers:
-# #   * ${CMAKE_CURRENT_BINARY_DIR}/bar_export.h -> <prefix>/include/bar_export.h
-# #   * ${CMAKE_CURRENT_BINARY_DIR}/baz_export.h -> <prefix>/include/baz_export.h
-# install(
-#     FILES
-#         "${CMAKE_CURRENT_BINARY_DIR}/bar_export.h"
-#         "${CMAKE_CURRENT_BINARY_DIR}/baz_export.h"
-#     DESTINATION "${include_install_dir}"
-# )
 
 # Config
 #   * <prefix>/lib/cmake/Foo/FooConfig.cmake
 #   * <prefix>/lib/cmake/Foo/FooConfigVersion.cmake
 install(
-    FILES "${project_config}" "${version_config}"
-    DESTINATION "${config_install_dir}"
+    FILES "${PROJECT_CONFIG}" "${VERSION_CONFIG}"
+    DESTINATION "${CONFIG_INSTALL_DIR}"
 )
 
 # Config
 #   * <prefix>/lib/cmake/Foo/FooTargets.cmake
 install(
-    EXPORT "${targets_export_name}"
+    EXPORT "${TARGETS_EXPORT_NAME}"
     NAMESPACE "${namespace}"
-    DESTINATION "${config_install_dir}"
+    DESTINATION "${CONFIG_INSTALL_DIR}"
 )
 ENDMACRO(_SETUP_PROJECT_PACKAGE_FINALIZE)
 
--- a/base.cmake
+++ b/base.cmake
@@ -56,6 +56,7 @@ INCLUDE(cmake/uninstall.cmake)
 INCLUDE(cmake/install-data.cmake)
 INCLUDE(cmake/release.cmake)
 INCLUDE(cmake/version.cmake)
+INCLUDE(cmake/package.cmake)
 
  # --------- #
  # Constants #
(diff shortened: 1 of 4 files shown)


haudren:
This is interesting but:
- Does it have any advantage over using `pkg-config` as is done currently ? Note that we use `pkg-config` also for the reason that we sometimes have to integrate with non-cmake build systems such as python's setuptools. Similarly, I wonder if this will play nicely with debian packaging...
- I see that the minimum cmake version in the project you linked is 3.0, can this work with 2.8 ? I know this is a hassle, but we have a bunch of Ubuntu 14.04 machines where 2.8.12 is still the default. At the very minimum, it should not break on those systems!

ahundt:
@haudren on 14.04 I believe the default is actually 2.8.7, so if you have 2.8.12 you've actually upgraded once already. [Here is a script you can use to update CMake](https://github.com/ahundt/robotics_setup/blob/master/cmake_source.sh).

Yes, there are a number of advantages, for a complete description I suggest reading the entire CMake page of [how to create relocatable packages](https://cmake.org/cmake/help/v3.6/manual/cmake-packages.7.html#creating-relocatable-packages). 

One such advantage is that as a user of a library can simply write the following:
```
find_project(RBDyn) # will be found automatically, and will create an imported target
add_library(mynewlib newlib.cpp)
target_link_libraries(mynewlib PUBLIC RBDyn) # will be an imported target
target_include_directories(mynewlib PUBLIC  $<TARGET_PROPERTY:RBDyn,INTERFACE_INCLUDE_DIRECTORIES>) # will include appropriate folders
```

ahundt:
here is a good guide that's easier to read: https://rix0r.nl/blog/2015/08/13/cmake-guide/

ahundt:
Also, this functionality should not interfere with pkg-config or debian packaging at all. 

The only possibility is that the scripts may need a small update to include the newly added cmake files to let `find_package()` work when installed via a debian package, since I wasn't familiar with those parts.

ahundt:
@haudren it looks like only sch-core has the cmake folder as a submodule, are the others (rbdyn, Tasks, SpaceVecAlg, Eigen3ToPython) using subtrees or is the directory simply committed?

haudren:
@ahundt : Unfortunately, as we have other users running Ubuntu 12.04 and hardware limitations, updating CMake is simply not an option at this time.

Yes, the projects that Joris Vaillant developped were stripped of their submodules to be replaced by direct commits of the directory as far as I remember.

I'll look in more details at the proposed changes later.

ahundt:
It is possible that this works with 2.8.12... but I haven't tested. Even when I'm using 12.04 I just install an updated cmake, though I understand the difficulty of getting lots of people to follow such instructions correctly to update and thus why you may choose to avoid it! 

Also one other thing is that project(ProjectName CXX) is missing the VERSION parameter, is there a good way to make the version be supplied correctly with the macros? I noticed that there is code to generate version numbers, so would like to add that in but I'm not sure of the details.

gergondet:
Hi @ahundt 

Thanks for the PR, I think this is a useful feature for packages that would like to rely on a pure CMake-based solution which seems to emerge as a de-factor standard in C++ development (e.g. vcpkg, conan and other such package managers). Having these files definitely goes toward that.

Before we merge, there is a few important remarks to address:

1. It absolutely should work on cmake 2.8.7 (it's the Ubuntu 12.04 default version which is still used/supported by most softwares that use the jrl-cmakemodules). We know that it can be updated but this is not a good option (e.g. for packaging) and the jrl-cmakemodules should work "out of the box" on distributions that we support.  I don't see anything in the changes that would not work on such version except for `include(CMakePackageConfigHelpers)` but this could be brought into jrl-cmakemodules (as we have done in the past for `GNUInstallDirs.cmake` if I recall correctly). If it is indeed not possible to make it work on such versions then the feature should be optional.
2. Retro-compatibility is a must. There is a lot more software that are using this submodule outside of this github group's projects. Updating the submodule in a repository should not require extra work to have the package compile. In particular, I am wondering what happens with the current PR if no EXPORT argument is provided in any INSTALL commands.
3. We have support to automatically generate different debug and non-debug targets (by setting the `PROJECT_DEBUG_POSTFIX` argument before `SETUP_PROJECT`. I'm guessing it works out of the box with a cmake file but please make sure that this is well supported (for example, RBDyn uses such a scheme)

For the code itself:
1. There is a lot of comments in `MACRO(_SETUP_PROJECT_PACKAGE_FINALIZE)` that seem pointless;
2. The start of `_SETUP_PROJECT_PACKAGE_INIT` and `_SETUP_PROJECT_PACKAGE_FINALIZE` setup the same variables
3. IMHO, the variables in these macros should be named `SETUP_ALL_CAPS_NAME` if you intend to use them in the user cmake scripts (e.g. `PROJECT_INCLUDE_INSTALL_DIR` rather than `include_install_dir`) as this will make it clearer that they are originating from jrl-cmakemodules.

For an overview of the `PROJECT_VERSION` computing scheme, you can take a look at #91 basically the usual scheme is to rely on tags and compute a minor version number based on the number of commit since this tag and the current commit hash.

PS: please refrain from submitting PR on other projects until this one is merged

ahundt:
> It absolutely should work on cmake 2.8.7 (it's the Ubuntu 12.04 default version which is still used/supported by most softwares that use the jrl-cmakemodules).

Hmm, 2.8.7 has serious problems that were fixed in 2.8.12, and additional changes in 3.x. 
I'm actually quite impressed that you support all the way back to 12.04.

I've dealt with these exact version conflicts (exactly ubuntu 12.04 + cmake 2.8.7, ubuntu 14.04 + cmake 2.8.12, and ubuntu 16.04 + cmake 3.5) with many other libraries and packages including OpenCV, ROS, PCL, other open source libraries and proprietary libraries. 
Usually the eventual decision on each of these projects which  is to provide cmake installer instructions for older versions either >2 or >4 years old for people that wish to backport so the project can move forward. I don't think 2.8.7 will support generator expressions like ` $<TARGET_PROPERTY:RBDyn,INTERFACE_INCLUDE_DIRECTORIES>`, though I believe 2.8.12 will.

One additional possibility aside from installing from source is using a [CMake 3.2 backport ppa for Ubuntu 12.04](http://askubuntu.com/questions/679299/is-there-a-ppa-offering-cmake-3-2-for-ubuntu-precise).

CMake itself provides excellent backwards compatibility so nothing should break for old CMake code running on new cmake versions.

> Retro-compatibility is a must. There is a lot more software that are using this submodule outside of this github group's projects. Updating the submodule in a repository should not require extra work to have the package compile. In particular, I am wondering what happens with the current PR if no EXPORT argument is provided in any INSTALL commands.

Yeah code currently breaks if there aren't any EXPORT targets. Will need a workaround or some sort of version policy to enable/disable the functionality. Perhaps simply checking if the targets list is empty before the install command would work?

> We have support to automatically generate different debug and non-debug targets (by setting the PROJECT_DEBUG_POSTFIX argument before SETUP_PROJECT. I'm guessing it works out of the box with a cmake file but please make sure that this is well supported (for example, RBDyn uses such a scheme)

Does this require two separate runs and installs of cmake? Could I get some example commands that reproduce the behavior to test? I think it should work, but I don't know how to use the functionality to verify.


> For the code itself:

These should all be fixed with 2700d88, with the exception of the `${namespace}` variable because it overlaps with `NAMESPACE`, could go with another suggestion. 

Side note: I suggest these handmade macro variables use a `JRL_` prefix or something, but that would be a breaking change.

> PS: please refrain from submitting PR on other projects until this one is merged

whoops... already done for all the ones I was going to do... sorry! I can update the same prs once this is complete.

Thanks for your replies and help!

ahundt:
> PROJECT_VERSION computing scheme, you can take a look at #91

Cool, looks interesting. [CMake 3.0+ has a new policy where the version can be set in the project() function](https://cmake.org/cmake/help/v3.0/command/project.html#command:project). Any chance of moving to a [semantic versioning scheme](http://semver.org/)?

This lets cmake automatically detect and warn about version differences.

jmirabel:
The very big advantage I see is the compatibility with Windows.

The `ADD_REQUIRED_DEPENDENCY` could be specialized on Windows to use CMake tools instead of pkg-config. The problem is that it should mimic the pkg-config behaviour regarding recursivity of the dependencies.

Do you think it is possible to integrate this in `ADD_REQUIRED_DEPENDENCY` for Windows ?

gergondet:
We use pkg-config on Windows for our packages so there is no reason to adapt `ADD_REQUIRED_DEPENDENCY` for this scenario.

ahundt:
What should be the next steps? 

Also, I'm not totally sure how to deal with a complete lack of export targets, maybe there is a way to detect if that's the case? My initial googling didn't turn up anything specific.

base.cmake
```diff
@@ -177,8 +208,79 @@ MACRO(SETUP_PROJECT)
...
+set(TARGETS_EXPORT_NAME "${PROJECT_NAME}Targets")
+set(namespace "${PROJECT_NAME}::")
+
+# Include module with fuction 'write_basic_package_version_file'
+include(CMakePackageConfigHelpers)
+
+# Configure '<PROJECT-NAME>ConfigVersion.cmake'
+# Note: PROJECT_VERSION is used as a VERSION
+write_basic_package_version_file(
+    "${VERSION_CONFIG}" COMPATIBILITY SameMajorVersion
```
gergondet: It's missing the `VERSION` argument

gergondet:
I've tested it in CMake 2.8.12 and it seems to work (except for the `VERSION` argument issue) when an `EXPORT` is provided

I've made some trials on my side as well and it seems like there is no way to find out whether an export target exists or not. It's actually possible to query a target's export name but it's not possible to access a list of a project's target. So far, I see no way to make this feature compatible with the existing cmake scripts that rely on jrl-cmakemodules without a modification on the users' side.

I suggest:
1. Keep the export initialization function in the `PROJECT_SETUP` macro
2. Move the finalize to a "public" macro (e.g. `PROJECT_EXPORT_FINALIZE`) that will be called explicitly by the users who want to generate a CMake export file.
3. Move both the init and export function to a dedicated file (e.g. `project-export.cmake`) to align with other features

I haven't had the time to test the validity of the files when generating them on a project where `PROJECT_DEBUG_POSTFIX` is set. To answer your question regarding the inner-working of this feature:
- with a single-build generator (e.g. Make or Ninja generator) you need to split your build directories according to the type of build you wish to have
- with a multi-build generator (e.g. Visual Studio, XCode and generally any IDE generator) then a single project is generated and the output name is properly set depending on the build setting.

Because the target has the same internal cmake name independently of the debug postfix I think it should work without problem through the export mechanism but it should be checked.

base.cmake
gergondet: It's not the case in [CMake 2.8](https://cmake.org/cmake/help/v2.8.8/cmake.html#module:CMakePackageConfigHelpers)

ahundt:
@gergondet That's a good suggestion, I'll enumerate some possible options including yours:


(1) Making `PROJECT_EXPORT_FINALIZE` a public macro users add as desired. This won't break
any existing code, and can easily be added for those who want it, but will require action by users.

(2) Moving the final step of `_PROJECT_EXPORT_FINALIZE` to user code:

```

# Config
#   * <prefix>/lib/cmake/Foo/FooTargets.cmake
install(
    EXPORT "${TARGETS_EXPORT_NAME}"
    NAMESPACE "${namespace}"
    DESTINATION "${CONFIG_INSTALL_DIR}"
)
```

Advising it be added when updating?

That may be doable without breaking anything that already exists but still add the basic config file. 


(3) A "dummy" project target could be added with the project name and perhaps the uninstall target. This could satisfy the source of the error without breaking anything. Or perhaps it can be safely assumed that all (or the vast majority) of the use cases you guys have include a target with the name of the project.

gergondet:
Of the 3 options, I'd rather go with (1) or (3). The benefit of (1) over (2) is that the behavior can be changed in the future without much trouble. (1) is also obviously the less disruptive.

I like (3) as well, I tested it using the following code and it works well:
```cmake
add_executable("${PROJECT_NAME}_EXPORT_DUMMY_TARGET" "cmake/dummy.cpp")
install(TARGETS "${PROJECT_NAME}_EXPORT_DUMMY_TARGET"
        EXPORT "${TARGETS_EXPORT_NAME}"
        DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/cmake")
```

where `dummy.cpp` is an empty C++ executable (i.e. `int main() { return 0; }`)

The problem I see is that it installs a supplementary file but that file is in the build folder so it should be treated as a build artifact akin to other files there. @bchretien I am correct in assuming that for Arch packages? @jmirabel would that be a problem for robotpkg?

jmirabel:
> would that be a problem for robotpkg?

I don't know, I'm not a robotpkg expert at all. @jcarpent may know.

> We use pkg-config on Windows for our packages so there is no reason to adapt ADD_REQUIRED_DEPENDENCY for this scenario.

Good to know that pkg-config works on Windows. I guess it is far from being in the Windows standard. So I would say it is not very satisfactory but it is off-topic here.

bchretien:
@gergondet if the file is copied to the build directory, there should not be any problem. Arch will only package the files installed in the install directory.

base.cmake
ahundt: so it should work if a VERSION argument is supplied? Can we even know that generally, some people may not have a version

gergondet: `PROJECT_VERSION` is always set by jrl-cmakemodules regardless of the users' inputs (at worst it will be `UNKNOWN` if there is no version provided otherwise and no git tag to find the version)

ahundt:
Went with option 1 since that is the most straightforward. Good to merge?

gergondet:
I'm alright with option 1.

Please move the macros to a dedicated file similarly to other macros used in `base.cmake`

ahundt:
any preferred name? package.cmake? did you want a different macro name as well?

gergondet:
I think `cmake-export.cmake` or `cmake-package.cmake` would be fine.

The name of the macro is fine.

ahundt:
I also made the VERSION change you asked for.

olivier-stasse:
I cannot say that cmake-package.cmake in a package using cmake for package management is very clear...

ahundt:
@olivier-stasse I named it simply `package.cmake`, and there is a comment description on the macro contained within. It could also be `package-config.cmake` or `packageConfig.cmake`.

jmirabel:
+1 for `package-config.cmake`.

ahundt:
@gergondet can we take the next steps on this?

gergondet:
Let's go with `package-config.cmake` then and please also move `_SETUP_PROJECT_PACKAGE_INIT` in that file.

After that I think we can merge.

ahundt:
@gergondet done, sorry for the delay. good to merge?

gergondet:
Looks good to me. Merging.

Thanks for the contribution :cake:

ahundt:
@gergondet note finishing integrating this will require small patches to the libraries that use it (sch-core, rbdyn, spacevecalg, tasks, etc). Should I create pull requests or would you be able to add that?

PR #107, opened 2016-12-15
possible typo.

proyan:
PKG_CONFIG_STRING is a variable in ADD_DOC_DEPENDENCY


jcarpent:
You're right. Thanks for the fix.

PR #110, opened 2017-03-08
Use a different strategy to repair cross links in doxygen

### LINKED ISSUE #104: Getting documentation cross-links to work
Hi,

Has anyone using the jrl-cmakemodules in combination with jrl-travis managed to get the cross-links between projects to work?

In my case, I've setup the proper TAGFILES entries in the [RBDyn](https://github.com/jrl-umi3218/RBDyn) repository but in the online [doc](https://jrl-umi3218.github.io/RBDyn) the links to the SpaceVecAlg library (e.g. click sva::RBInertiad on this [page](http://jrl-umi3218.github.io/RBDyn/doxygen/HEAD/a00002.html#aa7930e0bb454b55c51742130eb366aa8)) are definitely not correct. Part of the problem seems to be that doxygen does not generate any `installdox` script in my case but I'm not sure why...

I think this issue is somewhat related to #3 but I'm having trouble to get this feature to work even with a single TAGFILE entry. One possibility I see is to not rely on `installdox` and simply sed the local links to a generally correct github url (e.g. for a project organization/repo replace $install_dir/share/doc/repo/doxygen-html by organization.github.io/repo/doxygen/HEAD) based on the entries in a new argument to the `update-doxygen-doc.sh` script.

gergondet:
This PR fixes #104 

It accounts for the removal of installdox script in recent versions of doxygen. Instead, the local path to the doc is simply replaced by the URL deduced from the organization name + project name for those in the `GIT_DEPENDENCIES` variable.

### DIFF
--- a/github/update-doxygen-doc.sh
+++ b/github/update-doxygen-doc.sh
@@ -1,5 +1,5 @@
-#!/bin/sh
-# Copyright (C) 2008-2014 LAAS-CNRS, JRL AIST-CNRS.
+#!/bin/bash
+# Copyright (C) 2008-2017 LAAS-CNRS, JRL AIST-CNRS.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


github/update-doxygen-doc.sh
```diff
@@ -93,6 +59,51 @@ installdox_urls="$iu"
...
 
+# git_dependency_parsing
+# ----------------------
+#
+# From an entry in GIT_DEPENDENCIES variable set git_dep, git_dep_uri and
+# git_dep_branch in the environment
+# For example given the input "jrl-umi3218/jrl-travis" the following variables
+# are defined in the environment:
+# - git_dep jrl-umi3218/jrl-travis
+# - git_dep_uri git://github.com/jrl-umi3218/jrl-traviss
```
haudren: There is one `s` too much in here :smiling_imp:

github/update-doxygen-doc.sh
```diff
@@ -299,28 +310,20 @@ mkdir -p doxygen/${doc_version}
...
-    done
-    cd $tmp/project || abort "Failed to change directory."
-fi
+echo "* Fixing cross links..."
+cd "$tmp/project/doxygen/$doc_version"
+for package in ${GIT_DEPENDENCIES}
+do
+  git_dependency_parsing $package
+  if [ $git_dep_github = true ]; then
+    doxygen_url="https://${git_dep_organization}.github.io/${git_dep_project}/doxygen/HEAD/"
```
haudren: Do you think it would be possible to support other domains than `organization.github.io` ? I understand this will be the most common use-case, but if someone wishes to either host the documentation elsewhere (readthedocs) or simply to use the top-level URL (For example `robotpim.net` points to `roboptim.github.io`)

PR #111, opened 2017-05-02
Topic/find google benchmark

zbitouzakaria:
Added cmake script to find google's benchmark suite.

### DIFF
--- a/benchmark.cmake
+++ b/benchmark.cmake
@@ -1,0 +1,40 @@
+#Findbenchmark.cmake
+# - Try to find benchmark
+#
+# The following variables are optionally searched for defaults
+#  benchmark_ROOT_DIR:  Base directory where all benchmark components are found
+#
+# Once done this will define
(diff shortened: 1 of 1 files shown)


jmirabel:
In order to have some consistency between cmake helpers (boost, eigen...), I suggest to apply following changes:
- move current `benchmark.cmake` to `benchmark/FindBenchmark.cmake`
- add a file `benchmark.cmake`. This file should define a macro `SEARCH_FOR_BENCHMARK` adding `benchmark/` to the variable `CMAKE_MODULE_PATH` and calling `FIND_PACKAGE(Benchmark REQUIRED)

You could also put everything into a single macro `SEARCH_FOR_BENCHMARK`.

zbitouzakaria:
Hello jmirabel, 

Thank you for your advice. In the end we'd rather not add the library in this repository since it is used in only one of our projects.

PR #112, opened 2017-05-08
Create a dependency between template CSS creation and doc target

gergondet:
This PR is a follow-up to #72 

This makes doc dependent on the generation of doxygen template CSS. Otherwise, the custom command is not run and the generated documentation does not have its stylesheet.


jmirabel:
:+1:

PR #113, opened 2017-05-09
Topic/support version scripts

gergondet:
This PR adds a simple macro to add a version script (see: [doc](https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html)). These can be useful to hide undesired 3rd-party symbols.

The macro can be used whether the platform linker supports version scripts or not.


gergondet:
I believe this has no impact on existing projects, it just makes it easier to use version scripts in new projects.

I've kind of abandoned this PR back then because I didn't really need it and it was a bit pointless.

However, recently I've come across an issue that requires this feature. I am statically linking with the most recent version of the qhull library which is incompatible with one packaged in Ubuntu and which is linked by other components in the system (which I cannot change). By adding a version script to keep my own qhull symbols local I avoid the issue.

If there is no objection, I'll merge after updating the doc to the sphinx format.

Issue #44, opened 2014-04-24
Consistent testing between projects


bchretien:
cc @thomas-moulard @olivier-stasse @francois-keith 

In order to speed things up on Travis (since we're often reaching the deadly 50-min limit), I was considering changing [build_git_dependency()](https://github.com/jrl-umi3218/jrl-travis/blob/master/before_install#L50) in `jrl-travis` to disable test compilation for dependencies (unless explicitly asked).  This is something that was observed with RobOptim, so it may also concern other projects.
- 1°) We could use a simple check with an optional variable, e.g.:
  
  ``` cmake
  OPTION(BUILD_TESTS "Build test programs" OFF)
  IF(BUILD_TESTS)
      ADD_SUBDIRECTORY(tests)
  ENDIF(BUILD_TESTS)
  ```
- 2°) We could also use `EXCLUDE_FROM_ALL` with `ADD_EXECUTABLE` when adding tests, then add a dependency to trigger compilation when running `make test`, or expect a `make check` to be run before that (cf. [CMakeEmulateMakeCheck](http://www.cmake.org/Wiki/CMakeEmulateMakeCheck)).

I guess that no matter what we choose here, we need to make sure that all JRL/LAAS projects using the `jrl-umi3218` tools follow the same guidelines for their tests, so that Travis scripts can be modified accordingly. If this question was already asked and/or answered elsewhere, please let me know as well ;-)

PR #115, opened 2017-09-21
Add option COMPILE_UNIT_TESTS to exclude unit tests from target all

jmirabel:


florent-lamiraux:
+1

gergondet:
The patch is ok but there might be some issues depending on the context it would be used in.

See #44 for a related discussion. It also raise an issue with the solution you're proposing (namely, `make test` is broken as the test executable is not a dependency of the test target)

The common use case for such options is to disable tests in a CI environment. For this, the de-facto standard is the `DISABLE_TESTS` option which is passed to disable tests' builds on the project dependencies (see [this](https://github.com/jrl-umi3218/jrl-travis/blob/master/before_install#L144)]

jmirabel:
> See #44 for a related discussion. It also raise an issue with the solution you're proposing (namely, make test is broken as the test executable is not a dependency of the test target)

If I add the dependency you mentioned, would it be acceptable ? According #44, I do not think so.

> The common use case for such options is to disable tests in a CI environment.

I see other cases which can be interesting. Compiling unit-tests is useless for people interesting in trying a software as a user only. And for people developing code, it is *most of the time* not very useful as they do not run `make test` everywhere.

> The common use case for such options is to disable tests in a CI environment. For this, the de-facto standard is the DISABLE_TESTS option which is passed to disable tests' builds on the project dependencies (see this]

I will change the name of the variable.

gergondet:
> If I add the dependency you mentioned, would it be acceptable ? According #44, I do not think so.

If you have a solution for this that would be great (see below).

> I see other cases which can be interesting. Compiling unit-tests is useless for people interesting in trying a software as a user only. And for people developing code, it is most of the time not very useful as they do not run make test everywhere.

That sounds perfectly reasonable to me. I think making the unit tests dependent on the test target would make the option this PR introduces redundant as they won't be compiled unless the user wishes to run `make test` (which covers naive users, developers that are tired of recompiling the tests and CI).

Unfortunately I don't think this is as straight forward because test is a special target in cmake (see [this](https://cmake.org/pipermail/cmake/2009-January/026727.html)) but the following [SO post](https://stackoverflow.com/questions/733475/cmake-ctest-make-test-doesnt-build-tests) goes over some possible solutions.

jmirabel:
[This answer](https://stackoverflow.com/a/10824578) proposes a solution which is half way good. It would give something like:

- in the submodule:
```cmake
ADD_TEST(ctest_build_tests "${CMAKE_COMMAND}" --build ${CMAKE_BINARY_DIR} --target build_tests)
ADD_CUSTOM_TARGET(build_tests)
MACRO(ADD_UNIT_TEST NAME SOURCE)
  ADD_EXECUTABLE(${NAME} EXCLUDE_FROM_ALL ${SOURCE})
  ADD_DEPENDENCIES(build_tests ${NAME})
  ADD_TEST(${NAME} ${RUNTIME_OUTPUT_DIRECTORY}/${NAME})
  SET_TESTS_PROPERTIES ( ${NAME} PROPERTIES DEPENDS ctest_build_tests)
ENDMACRO()
```
- in developer CMakeLists.txt:
```cmake
ADD_UNIT_TEST(${NAME} ${NAME}.cc)
TARGET_LINK_LIBRARIES(${NAME} ...)
```

Then:
1. `make test` first run a test called `ctest_build_tests` and then run the actual tests.
2. `make build_tests` can be use to compile the tests.

I do no think this should be the default behavior but it could be enabled when `DISABLE_TESTS` is `ON`

jmirabel:
Mostly, the issues with this solution are:
- when compilation failed, the error message is not very clear (although `make build_tests` gives a proper error message).
- when compilation failed, it still tries to run the actual tests.

gergondet:
Ok, that's unfortunate but better than nothing I guess.

Then I suggest we adopt something similar to that solution, with:
- when `DISABLE_TESTS` is `OFF` (default) the tests are not `EXCLUDE_FROM_ALL`
- when `DISABLE_TESTS` is `ON` the tests are `EXCLUDE_FROM_ALL` but you still get the option to compile/run the tests with the caveats you mentioned

Bumping #66 to provide a comprehensive list of options/features provided by these scripts

base.cmake
```diff
@@ -162,6 +162,12 @@ MACRO(SETUP_PROJECT)
   ELSE(DEFINED INSTALL_DOCUMENTATION)
     SET(_INSTALL_DOC TRUE CACHE INTERNAL "")
   ENDIF(DEFINED INSTALL_DOCUMENTATION)
+  
+  # If variable COMPILE_UNIT_TESTS is TRUE,
+  # then directories added with the ADD_UNIT_TESTS_SUBDIRECTORY
+  # will be compiled with target all.
+  # COMPILE_UNIT_TESTS defaults to TRUE.
+  OPTION(COMPILE_UNIT_TESTS "Add unit tests directory to target all" ON)
```
gergondet: Shouldn't it be called `DISABLE_TESTS`?

gergondet:
Is there any reason why you didn't include `DISABLE_TESTS` as an option?

Otherwise, can this be merged?

jmirabel:
> Is there any reason why you didn't include DISABLE_TESTS as an option?

For the reasons mentioned in #44, i.e. to let people who want this as an option to explicitly say it. I hope this will avoid abuse of such an option.

Yes, it can be merged. I hope it won't break retro-compatibility. It shouldn't, except in some cases like target `build_tests` or macro `ADD_UNIT_TEST` are already taken and `cmake/test.cmake` is included (but this will be a cmake error which shouldn't be hard to fix).

gergondet:
Sounds reasonable.

`ADD_UNIT_TEST` is a fairly common macro name I'd say but cmake will simply overwrite the jrl-cmakemodules definition so that shouldn't break anything.

Merging now.

PR #117, opened 2017-10-04
Automatically link to doxytag file of dependencies found with pkg-config.

jmirabel:
This can be overwritten by adding to file `doc/Doxyfile.extra.in`:
```
TAGFILES =
```

When possible, relative paths are used. This makes the documentation easier to move.

The thing which is not very nice are:
- the iteration over `_PKG_CONFIG_REQUIRES` variable, which forces to manipulation the package name (as in macro `ADD_DEPENDENCY`). The same string manipulation is done twice.
- the use of `_PKG_CONFIG_DOXYGENDOCDIR`.


doxygen.cmake
```diff
@@ -184,6 +184,28 @@ MACRO(_SETUP_PROJECT_DOCUMENTATION_FINALIZE)
       SET(DOXYGEN_HEADER_NAME "header.html")
     ENDIF()
 
+    # Find doxytag files
+    # To ignore this list of tag files, add to doc/Doxyfile.extra.in
+    # TAGFILES =
+    STRING(REPLACE "," ";" PKG_REQUIRES "${_PKG_CONFIG_REQUIRES}")
```
gergondet: `PKG_CONFIG_REQUIRES` does not include the packages that were added as compile only dependencies. I don't think this feature is used a lot and afaik there is no variables that contains every `pkg-config` dependencies. Maybe this could be added.

doxygen.cmake
```diff
@@ -184,6 +184,28 @@ MACRO(_SETUP_PROJECT_DOCUMENTATION_FINALIZE)
...
+    FOREACH(PKG_CONFIG_STRING ${PKG_REQUIRES})
+      # Get the name of the package in the same format as in pkg-config.cmake.
+      STRING(REGEX MATCH "[^<>= ]+" LIBRARY_NAME "${PKG_CONFIG_STRING}")
+      STRING(REGEX REPLACE "[^a-zA-Z0-9]" "_" PREFIX "${LIBRARY_NAME}")
+      STRING(TOUPPER "${PREFIX}" "PREFIX")
```
gergondet: To simplify future maintenance, I'd rather have a macro/function that does this transformation in `pkg-config.cmake` and invoke it here.

doxygen.cmake
```diff
@@ -184,6 +184,28 @@ MACRO(_SETUP_PROJECT_DOCUMENTATION_FINALIZE)
...
+      # If DOXYGENDOCDIR is specified, add a doc path.
+      IF( DEFINED ${PREFIX}_DOXYGENDOCDIR
+          AND EXISTS ${${PREFIX}_DOXYGENDOCDIR}/${LIBRARY_NAME}.doxytag)
+        IF(DEFINED _PKG_CONFIG_DOXYGENDOCDIR)
+          FILE(RELATIVE_PATH DOXYTAG_FILE ${_PKG_CONFIG_DOXYGENDOCDIR} ${${PREFIX}_DOXYGENDOCDIR})
+        ELSE(DEFINED _PKG_CONFIG_DOXYGENDOCDIR)
```
gergondet: I'd rather we don't add repeated conditions in new code in the modules. They are the ghosts of cmake past 👻

doxygen.cmake
```diff
@@ -184,6 +184,28 @@ MACRO(_SETUP_PROJECT_DOCUMENTATION_FINALIZE)
...
+          SET(DOXYTAG_FILE ${${PREFIX}_DOXYGENDOCDIR})
+        ENDIF(DEFINED _PKG_CONFIG_DOXYGENDOCDIR)
+
+        SET(DOXYTAG_FILES "${DOXYTAG_FILES} ${${PREFIX}_DOXYGENDOCDIR}/${LIBRARY_NAME}.doxytag=${DOXYTAG_FILE}")
```
gergondet: I suggest the following rename (or something along those lines):

`DOXYTAG_FILES` -> `DOXYTAG_ENTRIES`

`DOXYTAG_FILE` -> `DEP_DOCDIR`

gergondet: You should also surround the entries with quotes here in case the prefix contains spaces (unless Doxygen does not accept that, in which case the users who have a space in their prefix will just have to live with their bad decision)

doxygen.cmake
jmirabel: I guess variable `_PKG_CONFIG_DOXYGENDOCDIR` is always defined when we reach this point.

doxygen.cmake
jmirabel: Does it make sense to link the doc to compile time dependencies ?

jmirabel:
> This could also be used in the SEARCH_FOR_EIGEN macro.

I do not understand what you meant.

pkg-config.cmake
```diff
@@ -208,6 +208,22 @@ MACRO(_SETUP_PROJECT_PKG_CONFIG_FINALIZE)
   ENDIF()
 ENDMACRO(_SETUP_PROJECT_PKG_CONFIG_FINALIZE)
 
+# _PARSE_PKG_CONFIG_STRING (PKG_CONFIG_STRING PREFIX)
```
gergondet: The doc here is inconsistent with the macro

pkg-config.cmake
```diff
@@ -208,6 +208,22 @@ MACRO(_SETUP_PROJECT_PKG_CONFIG_FINALIZE)
...
+# - _PKG_PREFIX_VAR <- MY_PACKAGE
+MACRO(_PARSE_PKG_CONFIG_STRING PKG_CONFIG_STRING _PKG_LIB_NAME_VAR _PKG_PREFIX_VAR)
+  # Retrieve the left part of the equation to get package name.
+  STRING(REGEX MATCH "[^<>= ]+" ${_PKG_LIB_NAME_VAR} "${PKG_CONFIG_STRING}")
+  # And transform it into a valid variable prefix.
+  # 1. replace invalid characters into underscores.
+  STRING(REGEX REPLACE "[^a-zA-Z0-9]" "_" ${_PKG_PREFIX_VAR} "${${_PKG_LIB_NAME_VAR}}")
+  # 2. make it uppercase.
+  STRING(TOUPPER "${${_PKG_PREFIX_VAR}}" ${_PKG_PREFIX_VAR})
+ENDMACRO(_PARSE_PKG_CONFIG_STRING PKG_CONFIG_STRING LIBRABRY_NAME VARIABLE_PREFIX)
```
gergondet: `ENDMACRO()`

doxygen.cmake
gergondet: You are right. This does not make much sense.

doxygen.cmake
```diff
@@ -184,6 +184,21 @@ MACRO(_SETUP_PROJECT_DOCUMENTATION_FINALIZE)
...
+    FOREACH(PKG_CONFIG_STRING ${PKG_REQUIRES})
+      _PARSE_PKG_CONFIG_STRING(${PKG_CONFIG_STRING} LIBRARY_NAME PREFIX)
+      # If DOXYGENDOCDIR is specified, add a doc path.
+      IF( DEFINED ${PREFIX}_DOXYGENDOCDIR
+          AND EXISTS ${${PREFIX}_DOXYGENDOCDIR}/${LIBRARY_NAME}.doxytag)
+        FILE(RELATIVE_PATH DEP_DOCDIR ${_PKG_CONFIG_DOXYGENDOCDIR} ${${PREFIX}_DOXYGENDOCDIR})
+
+        SET(DOXYTAG_ENTRIES "${DOXYTAG_ENTRIES} \"${${PREFIX}_DOXYGENDOCDIR}/${LIBRARY_NAME}.doxytag\"=\"${DEP_DOCDIR}\"")
+      ENDIF()
+    ENDFOREACH(PKG_CONFIG_STRING ${_PKG_CONFIG_REQUIRES})
```
gergondet: `ENDFOREACH()`

pkg-config.cmake
```diff
@@ -231,21 +247,9 @@ ENDMACRO(_SETUP_PROJECT_PKG_CONFIG_FINALIZE)
...
+  _PARSE_PKG_CONFIG_STRING ("${PKG_CONFIG_STRING}" LIBRARY_NAME PREFIX)
   IF(NOT ${PKG_CONFIG_DEBUG_STRING} STREQUAL "")
-    # Retrieve the left part of the equation to get package name.
-    STRING(REGEX MATCH "[^<>= ]+" LIBRARY_DEBUG_NAME "${PKG_CONFIG_DEBUG_STRING}")
-    # And transform it into a valid variable prefix.
-    # 1. replace invalid characters into underscores.
-    STRING(REGEX REPLACE "[^a-zA-Z0-9]" "_" ${PREFIX}_DEBUG "${LIBRARY_DEBUG_NAME}")
-    # 2. make it uppercase.
-    STRING(TOUPPER "${PREFIX}_DEBUG" "${PREFIX}_DEBUG")
+    _PARSE_PKG_CONFIG_STRING ("${PKG_CONFIG_DEBUG_STRING}" LIBRARY_DEBUG_NAME PREFIX_DEBUG)
```
gergondet: We want `${PREFIX}_DEBUG` here.

gergondet:
> > This could also be used in the SEARCH_FOR_EIGEN macro.
> 
> I do not understand what you meant.

I meant to add an Eigen entry into `DOXYTAG_ENTRIES` when `SEARCH_FOR_EIGEN` is called.

It should already work when Eigen is found through pkg-config as the macro adds eigen3 to `_PKG_CONFIG_REQUIRES` although I don't know if Eigen's doxytag filename follows the same convention as our packages (actually, the Ubuntu 14.04 package does not even include a doxytag file so this might be pointless...)

jmirabel:
I think I did all the changes which we mentioned, except for the doxytag of eigen, which I could not find (on my computer).

The dependencies added by macro `ADD_DOC_DEPENDENCY` are not handled by these changes. Should I do more changes in order to make it work ? (Mainly, adding a variable).

jmirabel:
Thanks

PR #118, opened 2017-10-12
Remove common path from complete path given as input

jcarpent:
This may be helpful to avoid long tree unfolding.


jcarpent:
@gergondet Thanks

PR #124, opened 2017-12-07
CMake support for multiple Python installations and versions.

ahundt:
CMake configuration can now specify the specific python executable, library, and version to utilize for when multiple versions of python are installed.

Based on my changes in https://github.com/bulletphysics/bullet3/pull/1340.

Notes on the new variables and the code to use are in the new [python/README.md](https://github.com/jrl-umi3218/jrl-cmakemodules/pull/124/files#diff-7141304209be3da2a29ef8b0b79886fd) file.


python/FindPythonLibs.cmake
```diff
@@ -1,40 +1,23 @@
-#.rst:
-# FindPythonLibs
-# --------------
-#
-# Find python libraries
-#
+# - Find python libraries
```
gergondet: You removed the sphinx doc here

python/FindPythonLibs.cmake
```diff
@@ -49,133 +32,209 @@
...
+# copyright note:
+
+# Copyright (c) 2007, Simon Edwards <simon@simonzone.com>
+# Redistribution and use is allowed according to the terms of the BSD license.
+# For details see the accompanying COPYING-CMAKE-SCRIPTS file.
+
+
+if(NOT DEFINED PYTHON_INCLUDE_DIR)
+  if(DEFINED PYTHON_INCLUDE_PATH)
+    # For backward compatibility, repect PYTHON_INCLUDE_PATH.
```
gergondet: respect

python/FindPythonLibs.cmake
```diff
@@ -49,133 +32,209 @@
...
-         REGEX "^#define[ \t]+PY_VERSION[ \t]+\"[^\"]+\"")
-    string(REGEX REPLACE "^#define[ \t]+PY_VERSION[ \t]+\"([^\"]+)\".*" "\\1"
-                         PYTHONLIBS_VERSION_STRING "${python_version_str}")
-    unset(python_version_str)
-  endif()
+  unset(_PYTHON_FIND_OTHER_VERSIONS)
+
+  message(STATUS "Looking for versions: ${_PYTHON_VERSIONS}")
+
+  FIND_FILE(_FIND_LIB_PYTHON_PY FindLibPython.py PATHS ${CMAKE_MODULE_PATH} ${CMAKE_ROOT}/Modules)
```
gergondet: Should be `"${CMAKE_ROOT}/Modules"` if there is space in cmake install path

python/FindPythonLibs.cmake
```diff
@@ -49,133 +32,209 @@
...
+    foreach(_CURRENT_PYTHON_NAME IN LISTS _PYTHON_NAMES)
+
+      unset(_PYTHON_EXECUTABLE CACHE)
+      find_program(_PYTHON_EXECUTABLE ${_CURRENT_PYTHON_NAME}
+        PATHS [HKEY_LOCAL_MACHINE\\SOFTWARE\\Python\\PythonCore\\${_CURRENT_VERSION}\\InstallPath])
+
+      if(_PYTHON_EXECUTABLE)
+
+        EXECUTE_PROCESS(
+          COMMAND ${_PYTHON_EXECUTABLE} "${_FIND_LIB_PYTHON_PY}"
```
gergondet: `"${_PYTHON_EXECUTABLE}"`

python/FindPythonLibs.cmake
```diff
@@ -194,33 +253,22 @@ set(PYTHON_DEBUG_LIBRARIES "${PYTHON_DEBUG_LIBRARY}")
...
-    set(PYTHONLIBS_VERSION_STRING "${PythonLibs_FIND_VERSION_MAJOR}")
-    if(PythonLibs_FIND_VERSION_COUNT GREATER 1)
-        set(PYTHONLIBS_VERSION_STRING "${PYTHONLIBS_VERSION_STRING}.${PythonLibs_FIND_VERSION_MINOR}")
-    endif(PythonLibs_FIND_VERSION_COUNT GREATER 1)
-    if(PythonLibs_FIND_VERSION_COUNT GREATER 2)
-        set(PYTHONLIBS_VERSION_STRING "${PYTHONLIBS_VERSION_STRING}.${PythonLibs_FIND_VERSION_TWEAK}")
-    endif(PythonLibs_FIND_VERSION_COUNT GREATER 2)
-endif()
-
+# include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)
```
gergondet: Why remove the hack and why this dead code?

python/SelectLibraryConfigurations.cmake
```diff
@@ -0,0 +1,70 @@
...
+    set( ${basename}_LIBRARIES "${${basename}_LIBRARY}" )
+
+    if( ${basename}_LIBRARY )
+        set( ${basename}_FOUND TRUE )
+    endif()
+
+    mark_as_advanced( ${basename}_LIBRARY_RELEASE
+        ${basename}_LIBRARY_DEBUG
+    )
+endmacro()
```
gergondet: What's the reason for including this file? AFAIK this is part of, at least, CMake 2.8

python/FindPythonLibs.cmake
```diff
@@ -194,33 +253,22 @@ set(PYTHON_DEBUG_LIBRARIES "${PYTHON_DEBUG_LIBRARY}")
 # what SELECT_LIBRARY_CONFIGURATIONS() expects.
 set(PYTHON_LIBRARY_DEBUG "${PYTHON_DEBUG_LIBRARY}")
 set(PYTHON_LIBRARY_RELEASE "${PYTHON_LIBRARY}")
-include(SelectLibraryConfigurations)
+include(${CMAKE_CURRENT_LIST_DIR}/SelectLibraryConfigurations.cmake)
```
gergondet: Why this change?

ahundt: Older versions of CMake don't have it

python/SelectLibraryConfigurations.cmake
ahundt: I don't remember the exact reason. I think there may have been a bug or it was missing from the 14.04 version... I mostly just brought some code over from the source at the top for pybullet then fixed a few bugs there, and now I'm bringing it over here. :-)

Here is some of the long meandering path this code has taken, after which it also visited pybullet and is now finally arriving here:
```
# Note by Nikolaus Demmel 28.03.2014: My contributions are licensend under the
# same as CMake (BSD). My adaptations are in part based
# https://github.com/qgis/QGIS/tree/master/cmake which has the following
```

ahundt:
The problem is you can't choose your python version and location. It will just use whichever one it finds first. If that happens to be a combination of python 2 executable and python 3 libraries it will try (and fail) to go with that. If you don't use the default system python (such as homebrewed python) it will also not work. 

In other words this makes sure you can choose which python to use and it verifies that the numpy and python library versions are consistent with your choice.

ahundt:
Here is an example of the errors I encounter when I try to install with python3, when `python` maps to `python2`:

```

+ pip3 install . --upgrade --user
Processing /Users/athundt/source/jrl-umi3218/sch-core-python
    Complete output from command python setup.py egg_info:

    Error compiling Cython file:
    ------------------------------------------------------------
    ...
    #
    # You should have received a copy of the GNU Lesser General Public License
    # along with sch-core-python.  If not, see <http://www.gnu.org/licenses/>.

    from eigen.c_eigen cimport Vector3d
    from sva.c_sva cimport *
    ^
    ------------------------------------------------------------

    sch/c_sch.pxd:19:0: 'sva/c_sva.pxd' not found

    Error compiling Cython file:
    ------------------------------------------------------------
    ...
    from sva.c_sva cimport *

    from c_sch cimport *

    cdef extern from "sch_wrapper.hpp" namespace "sch":
      void transform(S_Object & obj, const PTransformd&)
                                          ^
    ------------------------------------------------------------

    sch/c_sch_private.pxd:24:39: 'PTransformd' is not a type identifier

    Error compiling Cython file:
    ------------------------------------------------------------
    ...
    # along with sch-core-python.  If not, see <http://www.gnu.org/licenses/>.

    cimport c_sch
    cimport c_sch_private
    cimport eigen.eigen as eigen
    cimport sva.sva as sva
           ^
    ------------------------------------------------------------

    sch/sch.pyx:23:8: 'sva/sva.pxd' not found

    Error compiling Cython file:
    ------------------------------------------------------------
    ...
        if self.__own_impl and isinstance(self, S_Object):
          del self.impl
      def __cinit__(self):
        self.__own_impl = True
        self.impl = NULL
      def transform(self, sva.PTransformd trans):
                         ^
    ------------------------------------------------------------

    sch/sch.pyx:35:22: 'PTransformd' is not a type identifier

    Error compiling Cython file:
    ------------------------------------------------------------
    ...
          del self.impl
      def __cinit__(self):
        self.__own_impl = True
        self.impl = NULL
      def transform(self, sva.PTransformd trans):
        c_sch_private.transform(deref(self.impl), deref(trans.impl))
                                                 ^
    ------------------------------------------------------------

    sch/sch.pyx:36:46: Invalid operand type for '*' (Python object)
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/9f/jr_p974d3j318tmvrfkjl55w0000gp/T/pip-ge49lpql-build/setup.py", line 126, in <module>
        extensions = cythonize(extensions)
      File "/Users/athundt/Library/Python/3.6/lib/python/site-packages/Cython/Build/Dependencies.py", line 1039, in cythonize
        cythonize_one(*args)
      File "/Users/athundt/Library/Python/3.6/lib/python/site-packages/Cython/Build/Dependencies.py", line 1161, in cythonize_one
        raise CompileError(None, pyx_file)
    Cython.Compiler.Errors.CompileError: sch/sch.pyx
    sch/sch.pyx: cannot find cimported module 'sva.sva'
    ./sch/c_sch.pxd: cannot find cimported module 'sva.c_sva'
    ./sch/c_sch_private.pxd: cannot find cimported module 'sva.c_sva'
    Compiling sch/sch.pyx because it changed.
    [1/1] Cythonizing sch/sch.pyx

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/9f/jr_p974d3j318tmvrfkjl55w0000gp/T/pip-ge49lpql-build/
-> [1]
```

ahundt:
@gergondet I looked into the spot where you asked about the hack + dead code. With the new recommended search approach provided in the README.md this is no longer an issue because the python executable version is found first, which defines the complete search string and then the other version strings are based on that. Therefore, the hack would otherwise actively make mistakes and mismatch versions and thus should be removed.

python/SelectLibraryConfigurations.cmake
jmirabel: On my computer,
```bash
> apt-file list cmake | grep SelectLibraryConfigurations.cmake
cmake-data: /usr/share/cmake-2.8/Modules/SelectLibraryConfigurations.cmake
```

It seems to be in CMake 2.8 already. I didn't check the differences.

python/README.md
```diff
@@ -0,0 +1,60 @@
...
+```
+# Optional Python configuration
+# Will not probe environment for Python configuration (which can abort the
+# build process) unless you explicitly turn on PYTHON_BINDING.
+OPTION(PYTHON_BINDING "Set when you want to build PYTHON_BINDING (Python bindings for the library)" ON)
+IF(PYTHON_BINDING)
+	SET(PYTHON_BINDING_PYTHON_VERSION "" CACHE STRING "Python version PYTHON_BINDING will use.")
+	SET(Python_ADDITIONAL_VERSIONS 3 3.6 3.5 3.4 3.3 3.2 3.1 3.0 2.7 2.7.12 2.7.10 2.7.3 )
+	SET_PROPERTY(CACHE PYTHON_BINDING_PYTHON_VERSION PROPERTY STRINGS ${Python_ADDITIONAL_VERSIONS})
+	SET(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/build3/cmake ${CMAKE_MODULE_PATH})
```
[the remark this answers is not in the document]
ahundt: that's an example of a directory with cmake code in it. I guess in this case it would just be "${CMAKE_CURRENT_SOURCE_DIR}/cmake"

ahundt:
> I remember I encountered an issue some time ago when versions 2.7 and 3.4 of Python were both installed. The macro FINDPYTHON was updated #80 so that it is possible to specify an exact version of Python, like FINDPYTHON(2.7 EXACT).
>
> Is the current FINDPYTHON macro not working or do you add a functionality ?

I have python 2.7.10 and 2.7.13 installed, so the situation you described could mean I find 2.7.10 then link against 2.7.13 -> crash.

ahundt:
> It seems to be in CMake 2.8 already. I didn't check whether the file differs.

Honestly I don't know, I just took someone else's BSD licensed script, copied it into the projects where I use python, and fixed a few bugs. I didn't go into such details since the code worked. I assume they had it there because there exists some version of CMake where it was necessary. If you'd like I can remove that file.

Also here is my [robotics_tasks.sh](https://github.com/ahundt/robotics_setup/blob/master/robotics_tasks.sh) automated setup script to compile from source. Right now the [python3 install lines for sch-core](https://github.com/ahundt/robotics_setup/blob/master/robotics_tasks.sh#L205) don't work & I'd very much like to get these changes in so I can fix that!

python/SelectLibraryConfigurations.cmake
jmirabel: Except for comments, the two files are exactly the same for CMake 2.8 and CMake 3.2. I guess it can be removed.

python/FindNumPy.cmake
```diff
@@ -0,0 +1,59 @@
...
+        "(${PYTHON_NUMPY_VERSION_MAJOR} * 10000) + (${PYTHON_NUMPY_VERSION_MINOR} * 100) + ${PYTHON_NUMPY_VERSION_PATCH}")
+      string(REGEX REPLACE "\\\\" "/"  PYTHON_NUMPY_INCLUDE_DIR ${PYTHON_NUMPY_INCLUDE_DIR})
+    else()
+     unset(PYTHON_NUMPY_VERSION)
+     unset(PYTHON_NUMPY_INCLUDE_DIR)
+     message(STATUS "Requested NumPy version and include path, but got instead:\n${__output}\n")
+    endif()
+  endif()
+else()
+  message(STATUS "To find NumPy Python interpretor is required to be found.")
```
jmirabel: Shouldn't that be an error ?

jmirabel:
> I have python 2.7.10 and 2.7.13 installed, so the situation you described could mean I find the 2.7.10 executable then link against 2.7.13 -> crash. I also install sva on both 2.7 and 3, but python is python2, so install on python3 fails with the current code.

Yes, that's true, although I would say having Python 2.7.10 and 2.7.13 installed is a very awkward setting. Nevertheless, this change is interesting because it makes sure Python interpreter and Python libs are coherent.

Moreover, I think the macros `FINDPYTHON` and `FIND_NUMPY` should be updated.

python/FindNumPy.cmake
ahundt: I don't think so unless the user said it should be required

ahundt:
> Moreover, I think the macros FINDPYTHON and FIND_NUMPY should be updated.

How so? I was just hoping to get to a minimal working state, rather than write the ultimate python finder. :-)

jmirabel:
I understand your point. I just think it is not a good idea to have two ways to find the same library. Currently, the macros `FINDPYTHON` and `FIND_NUMPY` (obviously, the names are coherent) do not use the changes you propose (at least for CMake >= 3.2).

ahundt:
> I understand your point. I just think it is not a good idea to have two ways to find the same library. Currently, the macros FINDPYTHON and FIND_NUMPY (obviously, the names are coherent) do not use the changes you propose (at least for CMake >= 3.2).

Sorry in what context do you mean this?

I did create an issue on cmake itself, but it will be some time before one can define a minimum release version incorporating the fixes if/when they come around:
https://gitlab.kitware.com/cmake/cmake/issues/17546

ahundt:
Apparently there might be new find scripts that work with specific python versions, could those be used?

ahundt:
I'm still not able to compile the libraries in a script without the modifications in this pull request, I get the exact same error as before.

python/FindPythonLibs.cmake
ahundt: a correct implementation is provided in the directory now

ahundt:
I think this is a more minimal fix https://github.com/jrl-umi3218/sch-core/pull/20

gergondet:
Sorry for the lack of update on this. I think this could be merged. However, there's been some changes in the `FindPythonLibs.cmake` script since the PR, in particular, it's using `CMAKE_FIND_FRAMEWORK` for Apple. If you could include these changes in your PR that'd be great.

jcarpent:
Closing this PR as it seems #438 is proposing an alternative version compliant with the current module.

PR #128, opened 2018-05-01
Pull request to allow various python interpreter versions

olivier-stasse:
The current version of python detection is allowing only the default value of libboost-python library.
For instance, if ones has libboost-python installed on a xenial distribution it will links by default to:
```
/usr/lib/x86_64-linux-gnu/libboost_python.so
```
The problem is that:
```
/usr/lib/x86_64-linux-gnu/libboost_python.so -> libboost_python-py27.so
```
However the package libboost-python1.58-dev provides:
```
/usr/lib/x86_64-linux-gnu/libboost_python-py27.so
/usr/lib/x86_64-linux-gnu/libboost_python-py35.so
/usr/lib/x86_64-linux-gnu/libboost_python.a
/usr/lib/x86_64-linux-gnu/libboost_python.so
```
In order to link to 
```
/usr/lib/x86_64-linux-gnu/libboost_python-py35.so
```
We have to find to the cmake variable:
```
Boost_PYTHON-PY35_LIBRARY
```
This pull request implements this by specifying 
```
FIND_PYTHON(python-py35)
```


gergondet:
Looks good but I think you mean that it is used by specified `python-py35` instead of `python` in the boost components rather than `FIND_PYTHON(python-py35)` (since the PR changes nothing except adding messages in that macro)

Is it correct?

jmirabel:
Does this ensure consistency between the version of Python found with `FINDPYTHON` and the version of Boost Python found with `SEARCH_FOR_BOOST` ?

jmirabel:
I mean that, with `FINDPYTHON`, you can do something like:
```cmake
FINDPYTHON(3.5 EXACT REQUIRED)
```

Will then `SEARCH_FOR_BOOST` be consistent with this Python version ? This would mean the order of calls matters...

olivier-stasse:
Tests are passing with 
```
FINDPYTHON(3.5 EXACT REQUIRED)
```
and
```
FINDPYTHON(3.5 EXACT REQUIRED)
```
for eigenpy.
You can find the two repos with the uncorrect cmake submodule here:
```
https://github.com/olivier-stasse/eigenpy/tree/develpy35
```
and
```
https://github.com/olivier-stasse/eigenpy/tree/develpy27
```
Sorry no travis has been setup for eigenpy...

olivier-stasse:
Are you ok with this PR ?

olivier-stasse:
@jcarpent Cool sorry for overlooking this.

olivier-stasse:
Sorry guys in order to move on, I will merge the pull request and do a reverse in case of problem.

PR #130, opened 2018-05-03
[Boost] Handle recent version of Boost

jcarpent:
This removes a warning in recent version of cmake >= 3.11.1.
This PR is enforced version of #128.

@olivier-stasse Can you check on your side that this PR is compatible with your recent modifications?


boost.cmake
```diff
@@ -41,11 +41,53 @@ MACRO(SEARCH_FOR_BOOST)
...
+      IF(NOT PYTHONLIBS_FOUND)
+        MESSAGE(FATAL_ERROR "PYTHON has not been found. Please first check for Python.")   
+      ENDIF(NOT PYTHONLIBS_FOUND)
+
+      LIST(REMOVE_AT BOOST_COMPONENTS_ ${PYTHON_IN_BOOST_COMPONENTS})
+      SET(BOOST_PYTHON_COMPONENT "python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
+      LIST(APPEND BOOST_COMPONENTS_ ${BOOST_PYTHON_COMPONENT})
+    ENDIF()
+  ELSE(Boost_VERSION VERSION_GREATER "1.67.0")
+    IF(${PYTHON_VERSION_MAJOR} EQUAL 3)
```
gergondet: I think this is not defined if you haven't searched for Python before so you should probably check for `PYTHONLIBS_FOUND` first.

boost.cmake
```diff
@@ -41,11 +41,53 @@ MACRO(SEARCH_FOR_BOOST)
...
 
+  # Check if python is in the list and adjust the version according to the current Python version.
+  # This is made mandatory if for Boost version greater than 1.67.0
+  IF(Boost_VERSION VERSION_GREATER "1.67.0")
+    LIST(FIND BOOST_COMPONENTS python PYTHON_IN_BOOST_COMPONENTS)
+    SET(BOOST_COMPONENTS_ ${BOOST_COMPONENTS})
+    IF(${PYTHON_IN_BOOST_COMPONENTS} GREATER -1)
+      # Check if Python has been found
+      IF(NOT PYTHONLIBS_FOUND)
+        MESSAGE(FATAL_ERROR "PYTHON has not been found. Please first check for Python.")
```
gergondet: So the intent here is to imply that `FindPython` is called before `Search_For_Boost` if the former contains the python component right?

I think the message could contain this as a hint (as this is potentially breaking existing scripts)

boost.cmake
```diff
@@ -41,11 +41,53 @@ MACRO(SEARCH_FOR_BOOST)
...
+
+        IF(${UNIX})
+          SET(BOOST_PYTHON_COMPONENT "python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
+        ELSE(${UNIX})
+          SET(BOOST_PYTHON_COMPONENT "python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
+        ENDIF(${UNIX})
+        LIST(APPEND BOOST_COMPONENTS_ ${BOOST_PYTHON_COMPONENT})
+      ENDIF(${PYTHON_IN_BOOST_COMPONENTS} GREATER -1)
+    ELSE(${PYTHON_VERSION_MAJOR} EQUAL 3) 
+      SET(BOOST_COMPONENTS_ ${BOOST_COMPONENTS})
```
gergondet: This is assuming the default for `libboost-python.so` is going to remain 2.7 as default. I'm guessing this will change sooner than later with distributions shifting their default `python` to Python 3.

Unless the default is decided by Boost and not the distribution. In that case, the assumption is true until 1.67 right?

boost.cmake
jcarpent: You're right. I will do the modification.

boost.cmake
jcarpent: Yes, indeed.
I will do also the modification.

boost.cmake
jcarpent: This is what I read in the new FindBoost files concerning Boost 1.67.0
Anyway, the standard procedure should remain valid for Python 2.7 and Boost < 1.67.0.

jcarpent:
@gergondet Thank you for the very helpful comments

jcarpent:
Concerning #128, this PR is complementary, allowing to select the right version of boost-python component according to the Python version which has been found by FindPython macro.

olivier-stasse:
Sorry for being late in the comments. 

Actually Anthony Mallet proposed a different patch which is working with different version of python without to resort to all the tests introduced by this PR.  
The problem is that it is very likely that this wil only work on UNIX.

So in any case, I think IMHO it is currently the only proper way to handle the matter at this level.
So I agree with the PR (even so it is late) thanks again for your help...

PR #134, opened 2018-05-11
[Boost] Fix bug in checking Python

jcarpent:
I introduced a bug in #130 when checking Python even if the BOOST_COMPONENTS variable does not contain python.


jcarpent:
@florent-lamiraux This fixes your issue

PR #135, opened 2018-05-13
[Boost] Extract useful lines from IF (APPLE).

florent-lamiraux:


jcarpent:
Sorry to have introduced this bug. This is fine for me.

PR #137, opened 2018-05-30
add optional arg to ADD_PYTHON_UNIT_TEST to update PYTHONPATH

nim65s:
Hi,

This PR adds a `MODULE_PATH` optionnal argument to `ADD_PYTHON_UNIT_TEST`, which is useful to launch `.py` tests that need to import a compiled module in `CMAKE_BINARY_DIR`.

It also ensures that `SOURCE` is an absolute path, by setting it relative to `PROJECT_SOURCE_DIR`.


test.cmake
```diff
@@ -50,14 +50,20 @@ MACRO(ADD_UNIT_TEST NAME SOURCE)
...
+#   `SOURCE` is relative to `PROJECT_SOURCE_DIR`
 #
 #   .. note:: :command:`FINDPYTHON` should have been called first.
 #
 MACRO(ADD_PYTHON_UNIT_TEST NAME SOURCE)
-  ADD_TEST(${NAME} ${PYTHON_EXECUTABLE} "${SOURCE}")
+  SET(MODULE_PATH "${ARGN}")  # ARGN is not a variable
+  ADD_TEST(${NAME} ${PYTHON_EXECUTABLE} "${PROJECT_SOURCE_DIR}/${SOURCE}")
+  IF(MODULE_PATH)
+    SET_PROPERTY(TEST ${NAME} PROPERTY ENVIRONMENT "PYTHONPATH=${CMAKE_BINARY_DIR}/${MODULE_PATH}")
```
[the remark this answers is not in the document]
nim65s: It is empty by default, and IMO it should stay empty most of the time, but you're right, we should not disturb users who wants to set it, I'm fixing that.

test.cmake
```diff
@@ -50,14 +50,20 @@ MACRO(ADD_UNIT_TEST NAME SOURCE)
...
+#   `SOURCE` is relative to `PROJECT_SOURCE_DIR`
 #
 #   .. note:: :command:`FINDPYTHON` should have been called first.
 #
 MACRO(ADD_PYTHON_UNIT_TEST NAME SOURCE)
-  ADD_TEST(${NAME} ${PYTHON_EXECUTABLE} "${SOURCE}")
+  SET(MODULE_PATH "${ARGN}")  # ARGN is not a variable
+  ADD_TEST(${NAME} ${PYTHON_EXECUTABLE} "${PROJECT_SOURCE_DIR}/${SOURCE}")
+  IF(MODULE_PATH)
+    SET_PROPERTY(TEST ${NAME} PROPERTY ENVIRONMENT "PYTHONPATH=${CMAKE_BINARY_DIR}/${MODULE_PATH}:$ENV{PYTHONPATH}")
```
jmirabel: A simple remark: when `$ENV{PYTHONPATH}` is empty, PYTHONPATH will be terminated by `some_path:` which might not be desired.

nim65s: The documentation says that «Non-existent directories are silently ignored.», and I thought that the directory with no name is non-existent… But it turns out an empty string is like `.` :/

jmirabel:
Sounds good to me.

nim65s:
Thanks for the review !

PR #138, opened 2018-06-07
Allow make test command even if DISABLE_TESTS is OFF

jcarpent:
We can now run 'make test': it will compile the tests followed by ctest command.


test.cmake
```diff
@@ -18,19 +18,19 @@
...
 IF(NOT DEFINED DISABLE_TESTS)
   SET(DISABLE_TESTS OFF)
 ENDIF(NOT DEFINED DISABLE_TESTS)
-IF(DISABLE_TESTS)
-  ADD_TEST(ctest_build_tests "${CMAKE_COMMAND}" --build ${CMAKE_BINARY_DIR} --target build_tests)
-  ADD_CUSTOM_TARGET(build_tests)
-ENDIF(DISABLE_TESTS)
+
+ADD_CUSTOM_TARGET(build_tests)
+ADD_TEST(ctest_build_tests "${CMAKE_COMMAND}" --build ${CMAKE_BINARY_DIR} --target build_tests)
```
jmirabel: I don't think we wan't to add test `ctest_build_tests` when DISABLE_TESTS is OFF. This target is a hack to bypass the lack of versatility of target test. I propose this instead:
```cmake
IF(DISABLE_TESTS)
   ADD_TEST(ctest_build_tests "${CMAKE_COMMAND}" --build ${CMAKE_BINARY_DIR} --target build_tests)
ENDIF()
ADD_CUSTOM_TARGET(build_tests)
```

test.cmake
```diff
@@ -40,13 +40,13 @@ ENDIF(DISABLE_TESTS)
...
-    ADD_TEST(${NAME} ${RUNTIME_OUTPUT_DIRECTORY}/${NAME})
-    SET_TESTS_PROPERTIES ( ${NAME} PROPERTIES DEPENDS ctest_build_tests)
   ELSE(DISABLE_TESTS)
     ADD_EXECUTABLE(${NAME} ${SOURCE})
-    ADD_TEST(${NAME} ${RUNTIME_OUTPUT_DIRECTORY}/${NAME})
   ENDIF(DISABLE_TESTS)
+
+  ADD_DEPENDENCIES(build_tests ${NAME})
+  ADD_TEST(${NAME} ${RUNTIME_OUTPUT_DIRECTORY}/${NAME})
+  SET_TESTS_PROPERTIES(${NAME} PROPERTIES DEPENDS ctest_build_tests)
```
jmirabel:
```cmake
IF(DISABLE_TESTS)
  ADD_EXECUTABLE(${NAME} EXCLUDE_FROM_ALL ${SOURCE})
  SET_TESTS_PROPERTIES ( ${NAME} PROPERTIES DEPENDS ctest_build_tests)
ELSE()
  ADD_EXECUTABLE(${NAME} ${SOURCE})
ENDIF()

ADD_DEPENDENCIES(build_tests ${NAME})
ADD_TEST(${NAME} ${RUNTIME_OUTPUT_DIRECTORY}/${NAME})

```

test.cmake
jcarpent: I realize that this is the only known solution to compile tests by calling `make test` without creating additional target like `make check`.

jmirabel: We should not modify the default behaviour of target test to avoid surprises to people. If you want to compile and run the test, you must run `make build_tests && make test`. This is the standard imposed by CMake and I think we should stick with it.

Now, coming to the hack of compiling the tests when a user does `make test` with DISABLE_TESTS enabled: the purpose is mostly to avoid stupid errors like "test failed" while you asked not to compile them but to run them...

jmirabel: So to conclude, when DISABLE_TESTS is disabled, `make test` should stick to CMake standard and not compile the unit tests.

jmirabel: Eventually, we can add a target `check`. I just prefer not to modify `test`.

jcarpent: I will do the modifications for supporting it.

PR #139, opened 2018-06-07
Add apple module to handle specific features for OS X

jcarpent:
This default policy is mostly standard for most of the projects on OSX systems


gergondet:
No problem with the PR but I was wondering why not include it by default?

I'm also wondering what is the point of `${isSystemdir}` since you set `CMAKE_INSTALL_RPATH` in both cases?

jcarpent:
I don't remember well, but certainly this IF condition is no more mandatory and can be removed.
We may include by default, but I prefer to not alter previous projects.

PR #143, opened 2018-07-13
make release: also run make dist

nim65s:
Hi,

I think that `make release` should also create the tarball, so I hereby suggest to add `make dist` inside it.

This is the purpose of this PR, plus updating the corresponding documentation.

Guilhem.


jcarpent:
I think this is a good point.
It is true that it is boring to have to write `make release` and `make dist`all the time.

PR #145, opened 2018-07-18
Improve packaging of projects

jcarpent:
This PR allows to handle the packaging of projects on APPLE systems.

### DIFF
--- a/dist.cmake
+++ b/dist.cmake
@@ -30,11 +30,17 @@ MACRO(_SETUP_PROJECT_DIST)
     FIND_PROGRAM(TAR tar)
     FIND_PROGRAM(GPG gpg)
 
+    IF(APPLE)
+      SET(GIT_ARCHIVE_ALL ${PROJECT_SOURCE_DIR}/cmake/git-archive-all.py)
+    ELSE(APPLE)
+      SET(GIT_ARCHIVE_ALL ${PROJECT_SOURCE_DIR}/cmake/git-archive-all.sh)
+    ENDIF(APPLE)
+
     # Use git-archive-all.sh to generate distributable source code
     ADD_CUSTOM_TARGET(distdir
       COMMAND
       rm -f /tmp/${PROJECT_NAME}.tar
-      && ${PROJECT_SOURCE_DIR}/cmake/git-archive-all.sh
+      && ${GIT_ARCHIVE_ALL}
       --prefix ${PROJECT_NAME}-${PROJECT_VERSION}/  ${PROJECT_NAME}.tar
       && cd ${CMAKE_BINARY_DIR}/
       && (test -d ${PROJECT_NAME}-${PROJECT_VERSION}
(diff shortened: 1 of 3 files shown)


dist.cmake
```diff
@@ -30,15 +30,21 @@ MACRO(_SETUP_PROJECT_DIST)
     FIND_PROGRAM(TAR tar)
     FIND_PROGRAM(GPG gpg)
 
+    IF(APPLE)
+      SET(GIT_ARCHIVE_ALL ${PROJECT_SOURCE_DIR}/cmake/git-archive-all.py)
+    ELSE(APPLE)
+      SET(GIT_ARCHIVE_ALL ${PROJECT_SOURCE_DIR}/cmake/git-archive-all.sh)
+    ENDIF(APPLE)
```
gergondet: Any reason not to use the Python one in both cases? (I'm guessing porting the shell version to OSX is more annoying that it's worth...)

distcheck.cmake
```diff
@@ -46,38 +46,8 @@ MACRO(DISTCHECK_SETUP)
...
-                   -e "'s|CMAKE_C_COMPILER-ADVANCED:INTERNAL=.\\+||g'"
-                   -e "'s|CMAKE_C_FLAGS-ADVANCED:INTERNAL=.\\+||g'"
-                   -e "'s|CMAKE_C_FLAGS_DEBUG-ADVANCED:INTERNAL=.\\+||g'"
-                   -e "'s|CMAKE_C_FLAGS_MINSIZEREL-ADVANCED:INTERNAL=.\\+||g'"
-                   -e "'s|CMAKE_C_FLAGS_RELEASE-ADVANCED:INTERNAL=.\\+||g'"
-                   -e "'s|CMAKE_C_FLAGS_RELWITHDEBINFO-ADVANCED:INTERNAL=.\\+||g'"
-                   -e "'s|CMAKE_DETERMINE_C_ABI_COMPILED:INTERNAL=.\\+||g'"
-                   _build/CMakeCache.txt
       && cd _build
+      && cmake ..
```
gergondet: I think this is very different with the original intent (and `cmake` is run anyway in the line below).

The benefit of the previous version is that it keeps the options that you enabled/disabled whereas this would erase any choice made.

Also, this still doesn't do anything on `APPLE` platform, is it desired?

git-archive-all.py
```diff
@@ -0,0 +1,505 @@
...
+        archiver.create(output_file_path, options.dry_run)
+    except Exception as e:
+        parser.exit(2, "{0}\n".format(e))
+
+    sys.exit(0)
+
+
+if __name__ == '__main__':
+    main()
+
```
gergondet: Could you give a link to where this script is hosted/developed originally? That would be useful to update the script in the future or forward bug fixes if the need arises for either of these.

distcheck.cmake
jcarpent: What I mean here is that the simple removing of cache data in the CMakeCache.txt is no more working.

jcarpent:
@gergondet I did the modification to remain compatible with previous behaviour.
By the way, I keep the two versions of git-archive-all to remain compatible with previous behaviour on LINUX systems.

jcarpent:
And thanks for your quick feedbacks!

gergondet:
Ok, thanks for the changes. I'll merge now!

PR #158, opened 2018-09-28
apple: change name of cmake function

jcarpent:


apple.cmake
```diff
@@ -15,11 +15,11 @@
 
 
 #.rst:
-# .. command:: APPLY_DEFAULT_APPLE_CONFIGURATION ()
+# .. command:: USE_DEFAULT_APPLE_CMAKE_CONFIGURATION ()
```
gergondet: There is a discrepancy here with the actual name of the macro.

I'm not sure which name you want to go with but I think it would be better to go with `USE_DEFAULT_APPLE_CMAKE_CONFIGURATION()` as `CMAKE_*` could look like an official CMake thing.

jcarpent: I do agree. Maybe it is better to close this PR and I will come back later if needed.

PR #162, opened 2018-10-18
Handle {MAJOR,MINOR,PATCH} version

jcarpent:
This PR also allows changing the header extension for a custom one (.hh, .hpp, etc)


version.cmake
```diff
@@ -102,31 +102,45 @@ MACRO(VERSION_COMPUTE)
     # Check if git describe worked and store the returned version number.
     IF(GIT_DESCRIBE_RESULT)
       MESSAGE(
-	"Warning: failed to compute the version number,"
-	" 'git describe' failed:\n"
-	"\t" ${GIT_DESCRIBE_ERROR})
+        "Warning: failed to compute the version number,"
+        " 'git describe' failed:\n"
+        "\t" ${GIT_DESCRIBE_ERROR})
```
jmirabel: Since you are at it, would you mind changing this to:
```cmake
      MESSAGE(AUTHOR_WARNING
        "failed to compute the version number,"
        " 'git describe' failed:\n"
        "\t" ${GIT_DESCRIBE_ERROR})
```

version.cmake
```diff
@@ -102,31 +102,45 @@ MACRO(VERSION_COMPUTE)
...
       STRING(REGEX REPLACE "^v" "" PROJECT_VERSION "${GIT_DESCRIBE_OUTPUT}")
       IF(NOT PROJECT_VERSION)
-	MESSAGE(
-	  "Warning: failed to compute the version number,"
-	  "'git describe' returned an empty string.")
-	SET(PROJECT_VERSION UNKNOWN)
+	      MESSAGE(
+	        "Warning: failed to compute the version number,"
+          "'git describe' returned an empty string.")
+        SET(PROJECT_VERSION UNKNOWN)
```
jmirabel:
```cmake
MESSAGE(AUTHOR_WARNING
  "failed to compute the version number,"
  "'git describe' returned an empty string.")
```

version.cmake
```diff
@@ -102,31 +102,45 @@ MACRO(VERSION_COMPUTE)
...
       ENDIF()
 
       # If there is a dash in the version number, it is an unstable release,
       # otherwise it is a stable release.
       # I.e. 1.0, 2, 0.1.3 are stable but 0.2.4-1-dg43 is unstable.
       STRING(REGEX MATCH "-" PROJECT_STABLE "${PROJECT_VERSION}")
       IF(NOT PROJECT_STABLE STREQUAL -)
-	SET(PROJECT_STABLE True)
+        SET(PROJECT_STABLE True)
```
jmirabel: This changes pollutes a bit the history.

jmirabel:
Thanks. This is a good idea.

As a remark, this may cause conflicts when updating the module in downstream packages.

jcarpent:
The default behavior remains the previous one (.hh extension).

jmirabel:
I meant that variable `PROJECT_MAJOR_VERSION` may already be set by the project manually. I do not have examples of packages that use this so it is not a hard requirement.

jmirabel:
I think it is very hard to know whether the developer manually added `#define PROJECT_MAJOR_VERSION` in his code. I think you should not worry.

header.cmake
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2008-2014 LAAS-CNRS, JRL AIST-CNRS.
+# Copyright (C) 2008-2014,2018 LAAS-CNRS, JRL AIST-CNRS, INRIA.
```
gergondet: Is it not more correct to update the copyright to `2008-2018` rather than having this gap? (note: I personally don't know)

header.cmake
```diff
@@ -71,6 +71,10 @@ MACRO(_SETUP_PROJECT_HEADER)
     STRING(REGEX REPLACE "[^a-zA-Z0-9]" "/" HEADER_DIR "${PROJECT_NAME}")
   ENDIF(DEFINED CUSTOM_HEADER_DIR)
 
+  IF(NOT DEFINED CUSTOM_HEADER_EXTENSION)
```
gergondet: 1. This should be documented as other global variables that have an influence on the modules behavior (see `base.cmake`)
2. I suggest renaming the variable `PROJECT_CUSTOM_HEADER_EXTENSION` for consistency with other such options

version.cmake
```diff
@@ -101,29 +101,47 @@ MACRO(VERSION_COMPUTE)
...
+        SET(PROJECT_VERSION_MAJOR UNKNOWN)
+        SET(PROJECT_VERSION_MINOR UNKNOWN)
+        SET(PROJECT_VERSION_PATCH UNKNOWN)
+      ELSE()
+        # Extract the version from PROJECT_VERSION
+        string(REPLACE "." ";" _PROJECT_VERSION_LIST "${PROJECT_VERSION}")
+        string(REPLACE "-" ";" _PROJECT_VERSION_LIST "${_PROJECT_VERSION_LIST}")
+        list(GET _PROJECT_VERSION_LIST 0 PROJECT_VERSION_MAJOR)
+        list(GET _PROJECT_VERSION_LIST 1 PROJECT_VERSION_MINOR)
+        list(GET _PROJECT_VERSION_LIST 2 PROJECT_VERSION_PATCH)
```
gergondet: What happens here if the version does not have a patch number (e.g. `1.0`)?

jcarpent: I just push a fix for that.

header.cmake
jcarpent: I applied the request modifications.

header.cmake
jcarpent: I've fixed it.

PR #164, opened 2018-10-23
Find Numpy for new boost libs

vsamy:
Some modification in boost.cmake to be able to find boost numpy libraries


nim65s:
@vsamy : thanks for the contribution, but for this solution lacks an issue. We can't check that your solution fix your issue, if we can't reproduce your issue first :)

So I'll close this, but if you have any issue with the current master, please open an issue, or re-open discussions on this PR.

PR #165, opened 2018-10-26
[Python] errors if python bin and libs versions differ 

jmirabel:


wxmerkt:
This appears to break on Ubuntu 18.04 Bionic, cf. #179

PR #167, opened 2018-10-29
[Boost-python] fix version detection on debian systems

nim65s:
Hi,

In debian based systems, instead of binding to libboost_python.so, which is a symlink to libboost_python-py27.so, I think that we should either use libboost_python-py27.so or libboost_python-py35.so.

In some other distributions, I have also found libboost_python2.so / libboost_python3.so, but I don't know if this is a safe enough default.


jcarpent:
Do you have troubles with the current version of the cmake modules?

nim65s:
Well… on 16.04, `cmake -DPYTHON_EXECUTABLE=/usr/bin/python3.5 ..` in eigenpy finds `-- Boost_PYTHON_LIBRARY: /usr/lib/x86_64-linux-gnu/libboost_python.so`, which is for python2.7

nim65s:
And, obviously, all tests are failing with:
```
1: ImportError: /usr/lib/x86_64-linux-gnu/libboost_python-py27.so.1.58.0: undefined symbol: PyClass_Type
```

This errors shows with the cmake submodule in master, but not in topic/boost_python-py35

jcarpent:
In #147, I have made some similar adjustments for APPLE systems. You can take inspiration from it to detect the right version for python3.

nim65s:
In #147 you removed that `SET(BOOST_PYTHON_COMPONENT "python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")` I just added back in this PR

jcarpent:
Yes, you're right.
But you have to be cleaner for `ELSEIF(EXISTS "/etc/debian_version") `. This is not robust.

jcarpent:
Ok. From my concern, I would add a macro IS_DEBIAN() and not just detect if with `ELSEIF(EXISTS "/etc/debian_version")`
I know that they are equivalent, but the day where Debian will change its configuration, this approach will fail.

nim65s:
So, basically, you propose to make a `IS_DEBIAN` macro, which will run `EXISTS "/etc/debian_version"` ?
I think we can do that if you want, but maybe in another PR

jcarpent:
Yes, that what I suggest.

jcarpent:
I let @bchretien doing the merge.

gergondet:
@nim65s If I'm not mistaken #168 actually superseeds this PR right? Can we close it?

(sorry for the noise, I posted this in #168 first ^^)

PR #168, opened 2018-10-29
WIP:  Adding CHECK_DEBIAN 

nim65s:
This PR follows discussions in #167, and should not be merged before it, and before a rebase.


base.cmake
```diff
@@ -254,3 +254,23 @@ MACRO(SETUP_PROJECT_FINALIZE)
...
+# ---------
+#
+# .. variable:: IS_DEBIAN
+#
+#   **REQUIRED**
+#
+#   A Boolean, indicating if the current system is Debian based
+#
+MACRO(CHECK_DEBIAN IS_DEBIAN)
+  SET(${IS_DEBIAN} EXISTS "/etc/debian_version")
```
jcarpent: You should PARENT_SCOPE here.

base.cmake
```diff
@@ -254,3 +254,23 @@ MACRO(SETUP_PROJECT_FINALIZE)
...
+# Variables
+# ---------
+#
+# .. variable:: IS_DEBIAN
+#
+#   **REQUIRED**
+#
+#   A Boolean, indicating if the current system is Debian based
+#
+MACRO(CHECK_DEBIAN IS_DEBIAN)
```
jcarpent: I think check_debian should not take any argument but should define a global variable.
You can have a look on https://gist.github.com/lbaehren/2413369

base.cmake
```diff
@@ -261,20 +261,9 @@ ENDMACRO(SETUP_PROJECT_FINALIZE)
...
-#   A Boolean, indicating if the current system is Debian based
-#
-MACRO(CHECK_DEBIAN IS_DEBIAN)
-  IF(EXISTS "/etc/debian_version")
-    SET(IS_DEBIAN TRUE)
-  ELSE(EXISTS "/etc/debian_version")
-    SET(IS_DEBIAN FALSE)
-  ENDIF(EXISTS "/etc/debian_version")
-ENDMACRO(CHECK_DEBIAN IS_DEBIAN)
+MACRO(CHECK_DEBIAN)
```
jcarpent: I think you have to add PARENT_SCOPE to a variable to make it public for the whole project.

nim65s: I tried this version on a debian system and on another unix system, and `DEBIAN_FOUND` was correct

nim65s: Moreover, `PARENT_SCOPE` is not documented in [`FIND_FILE`](https://cmake.org/cmake/help/latest/command/find_file.html)

jcarpent: PARENT_SCOPE is when you SET a variable.
Apparently, FIND_FILE is doing that for you.

jcarpent:
I let @bchretien doing the merge.

gergondet:
@nim65s If I'm not mistaken #168 actually superseeds this PR right? Can we close it?

PR #170, opened 2018-11-18
Add support for c++11

jcarpent:


cxx11.cmake
```diff
@@ -0,0 +1,35 @@
...
+# .. ifmode:: user
+#
+# .. command:: PROJECT_USE_CXX11
+#
+#    This macro set up the project to compile the whole project 
+#    with C++11 standards.  
+#
+MACRO(PROJECT_USE_CXX11)
+  IF(CMAKE_VERSION VERSION_LESS "3.1")
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
```
gergondet: We should check that the compiler supports the flag (that is the case for any recent enough Linux/OSX compiler but not MSVC in particular)

jcarpent: Sounds important indeed. I've just made an update of the PR to accound for it.

jcarpent:
@gergondet can you check this PR again?

cxx11.cmake
```diff
@@ -0,0 +1,47 @@
...
+#    with C++11 standards.  
+#
+MACRO(PROJECT_USE_CXX11)
+  CHECK_CXX_COMPILER_FLAG("-std=c++0x" COMPILER_SUPPORTS_CXX0X)
+  CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
+  IF(COMPILER_SUPPORTS_CXX0X OR COMPILER_SUPPORTS_CXX11)
+    IF(CMAKE_VERSION VERSION_LESS "3.1")
+      IF(COMPILER_SUPPORTS_CXX0X)
+        SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
+      ELSE()
```
gergondet: Should check `COMPILER_SUPPORT_CXX11` here and simply don't add a flag if the compiler supports neither (that's the case for MSVC)

jcarpent: This is what it is done but not explicitly for cxx11

gergondet: Sorry, I missed the initial condition check.

I would prefer this to be done only for CMake < 3.1, i.e. instead of the current:
```
CHECK_FLAGS()
IF(SUPPORT_FLAGS)
  IF(CMAKE_VERSION < 3.1)
    ...
  ELSE()
    ...
  ENDIF()
ENDIF()
```

I'd rather we do:
```
IF(CMAKE_VERSION < 3.1)
  CHECK_FLAGS()
  IF(SUPPORT)
    ..
  ELSE()
    // Maybe raise a cmake error here?
  ENDIF()
ELSE()
  ...
ENDIF()
```

cxx11.cmake
```diff
@@ -0,0 +1,47 @@
...
+        SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
+      ELSE()  
+        SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+      ENDIF()
+    ELSE()
+      SET(CMAKE_CXX_STANDARD 11)
+    ENDIF()
+    IF(APPLE)
+      IF(POLICY CMP0025)
+        CMAKE_POLICY(SET CMP0025 NEW)
```
gergondet: Is this related to C++11 support? Shouldn't that be part of `APPLY_DEFAULT_APPLE_CONFIGURATION()`?

jcarpent: This is very related to C++11 support. So, I think this is the right place.

cxx11.cmake
jcarpent: But we have to check the flags anyway, because `CMAKE_CXX_STANDARD` is not checking that for us.
I think it is better to provide a feedback to users at this stage.

gergondet: We could use `CXX_STANDARD_REQUIRED` in that case (see https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD_REQUIRED.html)

The whole problem is that `-std=c++0x` or `-std=c++11` is not supported by all compilers (most notably/importantly visual studio) although they might support C++11 or better. Also, checking compiler flags on Visual Studio/Windows is lengthy.

cxx11.cmake
gergondet: Ok, I don't mind keeping it in. Just for future reference, can you explain how this is related to C++11 support?  From my understanding, [CMP0025](https://cmake.org/cmake/help/v3.0/policy/CMP0025.html) is about the identifier of clang inside cmake scripts

cxx11.cmake
jcarpent: It has been added to 59bcf58.

jcarpent: As I'm not using Windows, I'm not able to adjust the right strategy.
I would prefer to fix in a future PR.

cxx11.cmake
jcarpent: It was for suppressing some boring messages I got when CMAKE_CXX_STANDARD was used. I can remove if it only concerns my implementation.

jcarpent:
@gergondet Do you approve this PR?

gergondet:
Yup, sorry it flew under my radar. Merging now.

PR #172, opened 2018-12-13
boost python lib name: use the same policy for Apple, NetBSD & ArchLinux

nim65s:
Hi,

This PR follows #168, and similarly adds a `CHECK_NETBSD` to find the correct name for the boost python component.


nim65s:
Hi,

ArchLinux has the same behavior,  and provides boost_python / boost_python3 / boost_python27 / boost_python37 but no boost_python2, so I updated the PR accordingly

boost.cmake
```diff
@@ -83,13 +85,13 @@ MACRO(SEARCH_FOR_BOOST)
       ENDIF(NOT PYTHONLIBS_FOUND)
       IF(${PYTHON_VERSION_MAJOR} EQUAL 3) 
         LIST(REMOVE_AT BOOST_COMPONENTS_ ${PYTHON_IN_BOOST_COMPONENTS})
-        IF(APPLE)
+        IF(APPLE OR NETBSD_FOUND)
```
gergondet: So, shouldn't that also apply to `ARCHLINUX_FOUND`?

nim65s: I missed it, but on the other hand, ArchLinux currently provides boost v1.68 so the ELSE(Boost_VERSION VERSION_GREATER 106699) can be ignored

gergondet: Ok, no problem then :) (sorry, I didn't check the bigger scope of this part of the script)

PR #173, opened 2018-12-20
Add files related to Humanoid Path Planner project

jmirabel:
This adds files which are specific to the HPP project.

If this is accepted, could I be granted permission to accept PRs for this repo ? I could then manage the PRs related to the HPP side.


jmirabel:
A comment on this PR:
I hesitated between doing this PR and creating a new repository. The first solution is simpler to maintain than another repository, at the cost of polluting a little this module.

gergondet:
I think this is ok. There is already a number of macros/files that are project specific.

> If this is accepted, could I be granted permission to accept PRs for this repo ? I could then manage the PRs related to the HPP side.

No problem.

However, you have introduced a temporary swap file in this PR. Please remove it (with a force push to hide it from the main repository history)

jmirabel:
> However, you have introduced a temporary swap file in this PR. Please remove it (with a force push to hide it from the main repository history)

Oops. Thanks for the insight.

PR #174, opened 2018-12-21
[Doc] Add compile time dependencies to doxytag entries.

jmirabel:


jmirabel:
I guess nobody objects to those changes. I merge.

PR #177, opened 2019-01-18
Get Python SOABI

nim65s:
Hi,

This PR allows users to use the variable `PYTHON_SOABI`, which is an empty string in Python 2, and an ABI tag (cf. [PEP 3149](https://www.python.org/dev/peps/pep-3149/)) prefixed with a dot in Python 3.

Using this, we can have .so files for multiple Python versions in the same prefix, which should ease the transition from Python 2 to Python 3 for end users, and also help on the packaging side.

For example, to create `mylib.so` for Python 2.7 and `my_lib.cpython-35m-x86_64-linux-gnu.so` for Python 3.5:
```
SET_TARGET_PROPERTIES(my_lib PROPERTIES OUTPUT_NAME "my_lib${PYTHON_SOABI}")
```

Then, to use such a library from another project:
```
TARGET_LINK_LIBRARIES(my_target "my_lib${PYTHON_SOABI}")
```


jcarpent:
@nim65s Can you amend the last commit with the previous one?

nim65s:
done

jcarpent:
Seems good to me now. I'm going to merge it now.

PR #180, opened 2019-01-28
[IDL] Add arguments to GENERATE_IDL_CPP

jmirabel:


jcarpent:
@jmirabel I think this PR is related to #81. Isn't it?

jmirabel:
Yes, it is. The thing I need it to expose omniidl arguments when this is requested.

PR #183, opened 2019-02-06
provide directly the INSTALL_DOCUMENTATION option

nim65s:
Hi,

This PR follows #99 and discussions in https://github.com/humanoid-path-planner/hpp-core/pull/145
I don't see the point in not providing the `OPTION` directly.


base.cmake
```diff
@@ -210,14 +210,8 @@ MACRO(SETUP_PROJECT)
...
-  
-  # If the variable INSTALL_DOCUMENTATION is not set, then the project will install the doc. Otherwise, 
-  # the doc will be installed if INSTALL_DOCUMENTATION is set to TRUE.
-  IF(DEFINED INSTALL_DOCUMENTATION)
-    SET(_INSTALL_DOC ${INSTALL_DOCUMENTATION} CACHE INTERNAL "")
-  ELSE(DEFINED INSTALL_DOCUMENTATION)
-    SET(_INSTALL_DOC TRUE CACHE INTERNAL "")
-  ENDIF(DEFINED INSTALL_DOCUMENTATION)
+
+  OPTION(INSTALL_DOCUMENTATION "Generate and install the documentation" ON)
```
jmirabel: Open question: should we keep *INSTALL_DOCUMENTATION* or should we use *PROJECT_*... ?

jcarpent:
I did that to remain compatible with previous projects and without imposing `INSTALL_DOCUMENTATOIN` in the CMake config.
But this solution is fine for me too.

base.cmake
nim65s: As we are already using this in many places, it would be easier for our projects not to change it. Apart from that, I have no preference towards one ore the other

jcarpent: Certainly `PROJECT_INSTALL_DOCUMENTATION`will never conflict with Cmake.

nim65s: On the other hand, if CMake ever provides an option for this, we'd better use it

jmirabel: > On the other hand, if CMake ever provides an option for this, we'd better use it

Yes, but that takes some time and the hypothetical CMake method might not be compatible with our method.

jcarpent: So, to conclude this conversation, the better is to keep `INSTALL_DOCUMENTATION`. We will change later if really needed.

jcarpent:
I'm going to merge it.

PR #185, opened 2019-02-07
Find CDD, CLP, TinyXML, glpk & qpOASES

nim65s:
Hi,

This PR adds helpers to find some libraries that are commonly used among Gepetto softwares, to avoid having to reinvent the wheel each and every time one of our projects needs one of those libs.

Example use, in a project where this repo is in the `cmake` directory, for the `MY_LIB` CMake target:
```
SET(CMAKE_MODULE_PATH "cmake")
FIND_PACKAGE("TinyXML" REQUIRED)
TARGET_INCLUDE_DIRECTORIES(${MY_LIB} PUBLIC ${TinyXML_INCLUDE_DIRS})
TARGET_LINK_LIBRARIES(${MY_LIB} ${TinyXML_LIBRARIES})
```


jcarpent:
I think all those FindPackageName should go into a dedicated repository inside the cmake modules.

nim65s:
Then we would have to use `SET(CMAKE_MODULE_PATH "cmake/finders")`. As this is less obvious, we would have to document that properly.

But why not.

nim65s:
Also, we would have to move `FindOpenRTM.cmake` to be consistent.  But I don't know who uses this file, or if anyone still use it. Anyway, moving / removing it might break a few things in unknown places.

jcarpent:
I don't think people are directly calling it. They are just defining the path to get access to it. 
@gergondet How do you proceed at JRL/IDH?

jmirabel:
I believe they use this macro, which is easy to update.

https://github.com/jrl-umi3218/jrl-cmakemodules/blob/bf10c6ca3565b96ab283edaa8f18b5b533e59da9/openrtm.cmake#L19

jmirabel:
Whether they are in *cmake* or in *cmake/dependencies* does not change the fact you must set the *CMAKE_MODULE_PATH*

jmirabel:
And, actually, those files are provided here although they should be provided elsewhere. I think it is a good practice to force users to modify the module path so they know that the package is found by this module and not by an officially maintained package finder.

gergondet:
`@gergondet How do you proceed at JRL/IDH?`

We don't use `FindOpenRTM.cmake`  in any current projects. In some projects we use a `Find*.cmake` script that we usually located in a different directory (e.g. `CMakeModules` and set `CMAKE_MODULE_PATH` accordingly).

> And, actually, those files are provided here although they should be provided elsewhere. I think it is a good practice to force users to modify the module path so they know that the package is found by this module and not by an officially maintained package finder.

I agree with that as well.

jcarpent:
Then following this discussion, @nim65s can you move all the Find*.cmake to a dedicated directory in order to clean the tree of files?
I will then merge your PR.

nim65s:
Agreed. But now, how do we call this directory ?

jmirabel:
Actually, I would go for a directory structure like:
- cmake
  - find-external
    - tinyxml
      - FindTinyXML.cmake
    - CDD
      - FindCDD.cmake

Add then:
```cmake
set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/cmake/find-external/tinyxml)
find_package (TinyXML)
```

This will make sure there are problem such that FindA.cmake and FindB.cmake are provided by this module but FindA.cmake is also officially provided.

jcarpent:
I do agree.

nim65s:
I find this overkill and too verbose,  especially when we need more than one of those libraries. 
But let's not fight about this, even if this is Friday :)

jcarpent:
But we need to be robust too.
I will merge this PR now. @nim65s Thank you for this contribution.

PR #196, opened 2019-03-05
Add .gitignore

jmirabel:
1d9aeca25e970d2d967fd5be0fb93fe961db121b introduced a Python script. When the Python script is used, the file is compiled and the pyc file makes the module dirty. This pollutes git status.

I didn't want to copy the file to the build directory because it seems overkill but I can do it if necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+hpp/idl/*.pyc


jcarpent:
With this solution, you cannot then have multiple builds. I mean, they will all share the same generated file, which might be not safe.

jmirabel:
The file is not generated by CMake. It will be the same for every build. It is a compiled version of the Python script. Even if it changes, Python will take care of removing it if the compiled version is not up to date.

jcarpent:
But this means that if you have a build with Python2 or Python3, it will not be the same.

jmirabel:
It won't be the same but the pyc file will be overwritten each time you need it (@nim65s can you confirm this ?).

nim65s:
pyc files names have changed between python 2 & python 3, ref [PEP 3147](https://www.python.org/dev/peps/pep-3147/), so you should not fear having a wrong one.

But they are generated files, so there is no point in versioning them, and therefore I usually put `*.pyc` in a global gitignore file.

jmirabel:
> therefore I usually put *.pyc in a global gitignore file.

Putting `*.pyc` is also good for me.

jcarpent:
maybe just a `hpp/idl/*.pyc` will be enough. In this way, we can still detect default in other directories.

jmirabel:
I updated the file accordingly.

jcarpent:
Looks good to me. I will merge it now.

PR #198, opened 2019-03-08
[IDL] Add arguments to omniidl macros.

jmirabel:


jmirabel:
Nobody objected so I merge this. Feel free to re-open if necessary.

PR #199, opened 2019-03-21
[pkg-config] NO_INCLUDE_SYSTEM option for PKG_CONFIG_USE_DEPENDENCY

### LINKED ISSUE #195: pkg-config : issue with SYSTEM in include_directories
When doing include_directories here
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/pkg-config.cmake#L764
and here
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/pkg-config.cmake#L766
the SYSTEM option is used.
It forces to uses -isystem instead of -I and change the order of inclusion with the header files.
It happens to be breaking the chaining of workspace in one of our ros package.
Do you mind if we remove them or is there any particular rational behing this choice ?

olivier-stasse:
This PR intends to fix #195.
It does not change the default behavior but add the possibility to the user to not force all the headers to be consider as system headers.
It is possible to do the following to avoid -isystem:
PKG_CONFIG_USE_DEPENDENCY(my_package the_dependency OPTIONAL NO_INCLUDE_SYSTEM)

IMHO forcing -isystem is nice to avoid warnings but might have dangerous side effect if not used with proper static code analysis.


pkg-config.cmake
```diff
@@ -892,8 +899,15 @@ ENDMACRO(BUILD_PREFIX_FOR_PKG)
 #       PKG_CONFIG_USE_DEPENDENCY(my-binary my-package)
 #
 MACRO(PKG_CONFIG_USE_DEPENDENCY TARGET DEPENDENCY)
+  SET(options OPTIONAL NO_INCLUDE_SYSTEM)
```
jmirabel: I think OPTIONAL is not expected here. It is understood by cmake_parse_arguments as an argument like any others.

olivier-stasse:
Sure I can.
It was just coming [from this part of the cmake documentation](https://cmake.org/cmake/help/v3.0/module/CMakeParseArguments.html):
```
Keywords terminate lists of values, e.g. if directly after a one_value_keyword another recognized 
keyword follows, this is interpreted as the beginning of the new option. E.g. my_install(TARGETS foo
 DESTINATION OPTIONAL) would result in MY_INSTALL_DESTINATION set to “OPTIONAL”, but
 MY_INSTALL_DESTINATION would be empty and MY_INSTALL_OPTIONAL would be set to TRUE
therefor.
```
This is interesting indeed only if we have several keywords which is not the case. 
Is my understanding correct ?
Or do we want force the user to understand that this an option ?
I have no preference.
If you confirm, I will remove it.

jmirabel:
I understand your point. However, I don't know any CMake commands working like that. Many have optional arguments and none have an argument to separate required and optional arguments.

I would be in favor of not putting the OPTIONAL keyword.

jmirabel:
Other opinions are welcome, of course.

olivier-stasse:
If I may and then after let us close the matter, install is documented this way:
https://cmake.org/cmake/help/latest/command/install.html

olivier-stasse:
Thx

pkg-config.cmake
```diff
@@ -761,9 +761,16 @@ MACRO(PKG_CONFIG_USE_LCOMPILE_DEPENDENCY TARGET PREFIX)
...
   # Include/libraries paths seems to be filtered on Linux, add paths
   # again.
-  INCLUDE_DIRECTORIES(SYSTEM ${${PREFIX}_INCLUDE_DIRS})
+  IF(NOT NO_INCLUDE_SYSTEM)
+    SET(INCLUDE_DIRECTORIES_VAR "SYSTEM ${${PREFIX}_INCLUDE_DIRS}")
+    SET(DEBUG_INCLUDE_DIRECTORIES_VAR "SYSTEM ${${PREFIX}_DEBUG_INCLUDE_DIRS}")
+  ELSE(NOT NO_INCLUDE_SYSTEM)
+    SET(INCLUDE_DIRECTORIES_VAR "${${PREFIX}_INCLUDE_DIRS}")
+    SET(DEBUG_INCLUDE_DIRECTORIES_VAR "${${PREFIX}_DEBUG_INCLUDE_DIRS}")    
+  ENDIF(NOT NO_INCLUDE_SYSTEM)
```
gergondet: I tested this on a simple project and this is broken. Here the CMakeLists.txt file where the call happens is `/tmp/foo` and this ends up in having `-I"/tmp/foo/SYSTEM /some/path"`

AFAIK, the proper way to do it is not to introduce an intermediary variable but call `include_directories` with/without the `SYSTEM` option.

gergondet:
I side with @jmirabel  regarding `OPTIONAL`. It serves no purpose here.

Regarding the documentation, the wording in CMake 3.0.4 documentation is a bit confusing (actually it even looks truncated) and I believe it's much clearer in [more recent documentation](https://cmake.org/cmake/help/v3.14/command/cmake_parse_arguments.html#command:cmake_parse_arguments).

>Keywords terminate lists of values, e.g. if directly after a one_value_keyword another recognized keyword follows, this is interpreted as the beginning of the new option. E.g. `my_install(TARGETS foo DESTINATION OPTIONAL)` would result in `MY_INSTALL_DESTINATION` set to `"OPTIONAL"`, but as `OPTIONAL` is a keyword itself `MY_INSTALL_DESTINATION` will be empty and `MY_INSTALL_OPTIONAL` will therefore be set to `TRUE`.

I am not sure what you meant by linking the install command documentation but `OPTIONAL` has a specific meaning related to the install command and is not used to specify optional arguments in some way.

PR #200, opened 2019-03-26
Extend package config support

gergondet:
This PR improves/extends package config support in jrl-cmakemodules by:
1. Introduce the `ADD_PROJECT_DEPENDENCY` macro that acts as a proxy for `find_package` and appends appropriate calls to `find_dependency` in the generated package config file
2. Adds support for defining `PROJECT_EXTRA_MACROS` to appends CMake macros to the generated package config file

This also fixes the issues introduced in #199 but I can do a separate PR for that


package-config.cmake
```diff
@@ -45,8 +45,22 @@ set(VERSION_CONFIG "${GENERATED_DIR}/${PROJECT_NAME}ConfigVersion.cmake")
...
+
+set(_PACKAGE_CONFIG_DEPENDENCIES "" CACHE INTERNAL "")
 ENDMACRO(_SETUP_PROJECT_PACKAGE_INIT)
 
+# ADD_PROJECT_DEPENDENCY
+# ----------------------
+#
+# This is a wrapper around find_package to add correct find_dependency calls in
+# the generated config script. All arguments are passed to find_package
+#
```
jmirabel: Can you make the doc in the good format for HTML doc ?

gergondet: Done.

I've added a page about this to the doc and I've also updated the documentation to reflect the addition of `NO_INCLUDE_SYSTEM`

jmirabel:
Thanks. That's a nice contribution.

PR #201, opened 2019-03-26
[pkg-config] Fix pb on SYSTEM

olivier-stasse:
Sorry the commit I proposed previously is generating a problem when using PKG_CONFIG_USE_DEPENDENCY.
This commit fixes it.
Should not be "SYSTEM ..."
but SYSTEM without quotes


olivier-stasse:
I close it because it is redundant with PR#200

PR #202, opened 2019-03-27
fix formatting errors

nim65s:
Hi,

I had the following error:
```
CMake Error: Error in cmake code at cmake/pkg-config.cmake:910:
Parse error.  Expected a newline, got identifier with text "SET".
CMake Error at cmake/base.cmake:98 (INCLUDE):
  include could not find load file:

    cmake/pkg-config.cmake
  CMakeLists.txt:3 (INCLUDE)
```

### DIFF
--- a/pkg-config.cmake
+++ b/pkg-config.cmake
@@ -907,8 +907,10 @@ ENDMACRO(BUILD_PREFIX_FOR_PKG)
 #                         this option changes this behaviour
 #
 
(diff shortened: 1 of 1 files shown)


gergondet:
Sorry about that, I must have broken it after adding the doc. Merging right now!

nim65s:
Thanks for the quick response :)

PR #203, opened 2019-03-28
[IDL] Always add prototype to documentation.

jmirabel:


jmirabel:
Thanks for the review.

PR #204, opened 2019-04-17
[pkg-config] customizable name for the .pc

nim65s:
Hi,

Following #177, I would need to generate different .pc files for the different installed version of one package.

Because if we want to install the same library for multiple python versions side-by-side, the `.pc` will probably contain things like `Cflags:  -isystem/usr/include/python3.5`, and `Libs: -l${PROJECT_NAME}{PYTHON_SOABI}`… So we need to be able to have eg. `py27-${PROJECT_NAME}.pc` & `py35-${PROJECT_NAME}.pc`.

In this PR, I propose a very generic way to allow that, and not something automatic to prepend `pyXY-` to this name, because we also have the same kind of issue with Qt 4 / 5.


nim65s:
After trying to use it, it doesn't fit my use case… I'll amend that, sorry.

nim65s:
That's better now :)

PR #205, opened 2019-04-29
pkg-config: recover full name of the main library

jcarpent:


jmirabel:
This is a hack of the macro which may have surprising results...

If someone wants a different .pc filename, I think he/she should be specified it manually rather than through the library output name.

Moreover, I understand the reason why you want to have several .pc file for the same library. I think what you aim at is much easier to achieve with CMake config files, with components, rather than with a bunch of .pc files.

jcarpent:
@jmirabel Thanks for your quick feedback. I realized I made a mistake with the previous commit. I've forced push in order to directly get the name of the output library.

jmirabel:
This makes more sense. I don't really like the behavior of this macro but your changes do not change what I dislike so I am fine with this.

jcarpent:
@jmirabel Sounds good to me. Thanks a lot for your feedback.

pkg-config.cmake
```diff
@@ -693,12 +693,13 @@ MACRO(PKG_CONFIG_APPEND_LIBS LIBS)
       # Check if this project is building this library
       IF(TARGET ${LIB})
         # Single build type generator
+        GET_TARGET_PROPERTY(OUTPUT_LIB_NAME ${PROJECT_NAME} OUTPUT_NAME)
```
jmirabel: I get issues with this, with CMake 3.5.1.

1. Shouldn't this use `${LIB}` instead of `${PROJECT_NAME}` ?
2. OUTPUT_LIB_NAME is set to NOTFOUND. The target does exist and generates a library.
  - Command `GET_PROPERTY(OUTPUT_LIB_NAME_IS_SET TARGET ${PROJECT_NAME} PROPERTY OUTPUT_NAME SET)` tells me that the property is not set.
  - Command `GET_PROPERTY(OUTPUT_LIB_NAME_IS_SET TARGET ${PROJECT_NAME} PROPERTY OUTPUT_NAME FULL_DOCS)` tells me that the property is not found.

Changing this line to
```cmake
        GET_PROPERTY(OUTPUT_LIB_NAME_SET TARGET ${PROJECT_NAME} PROPERTY OUTPUT_NAME_Debug SET)
        IF(OUTPUT_LIB_NAME_SET)
          GET_TARGET_PROPERTY(OUTPUT_LIB_NAME ${LIB} OUTPUT_NAME) 
        ELSE(OUTPUT_LIB_NAME_SET)
          SET(OUTPUT_LIB_NAME ${LIB})
        ENDIF(OUTPUT_LIB_NAME_SET)
```
but I would prefer to understand why OUTPUT_NAME is not set...

jcarpent: What is `${LIB}$? Who sets this variable?

The idea of this PR is that for some projects, the output name of the main library is changed (for example to add cython in the name for Python libraries). In this case, one has to take care of reporting this change of naming in the generation of the *.pc file.

jcarpent: The solution you are suggesting is fine for me.

gergondet: I think `OUTPUT_NAME` is only set when it is set explicitly through a `set_target_properties` call (maybe not in later versions of CMake). imho `$<TARGET_FILE:${TGT}>` is a much more reliable way to get the target name but it may require to generate the .pc rather than merely configuring it.

jmirabel: > What is `${LIB}$? Who sets this variable?

It is defined a few line above by `FOREACH(LIB ${LIBS})` where `LIBS` is the argument of the function.

jmirabel: In the current version, the macro argument isn't used...

jmirabel: @gergondet How would you obtain `$<TARGET_FILE:${TGT}>` ?

Generator expressions seem to be allowed only in a subset of command ([source](https://cmake.org/cmake/help/v3.3/manual/cmake-generator-expressions.7.html))

gergondet: The subset is actually much larger than the documentation might suggest here. In particular, we could use [`file(GENERATE...)`](https://cmake.org/cmake/help/v3.14/command/file.html#generate) in addition of `configure_file` to generate the .pc with full names

jcarpent: > In the current version, the macro argument isn't used...

This is a mistake indeed.

jmirabel: Generator expression would be good to generate the file itself. But
- I couldn't find a relevant *informational expression* that would give us straightaway the pkg-config file name. `$<TARGET_FILE_NAME:tgt>` is noisy. We would have to remove *lib* and *.so*, at least. Is it the strategy you have in mind ?
- I couldn't get the generator expression to be evaluated in a string. I tried with `STRING(GENEX_STRIP "$<TARGET_FILE_NAME:${LIB}>" NAME)` as the documentation of STRING suggests, but it didn't work...

Is it very bad to rely on the person who writes the CMakeLists for this ?
We may be able to derive a smart strategy. However, I am sure that, one day, someone will come with a case which is not correctly handled. I would prefer a *simple macro* that do not guess much, and, eventually, provide a low-level macro for the developers who want to implement, in their CMakeLists.txt, a more advanced behavior.

gergondet: > I couldn't find a relevant informational expression that would give us straightaway the pkg-config file name. $<TARGET_FILE_NAME:tgt> is noisy. We would have to remove lib and .so, at least. Is it the strategy you have in mind ?

Yes indeed, I got confused about the goal here and I thought you were trying to get the full path. It seems that getting the `OUTPUT_NAME` property and falling back to the target name when it is not set is the correct way to get it.

> I couldn't get the generator expression to be evaluated in a string. I tried with `STRING(GENEX_STRIP "$<TARGET_FILE_NAME:${LIB}>" NAME)` as the documentation of STRING suggests, but it didn't work...

My understanding is that this is meant to remove generator expressions from strings, so the output would be `TARGET_FILE_NAME:${LIB}`. The trick with generator expressions is that they happen at generation time and not configuration time in the cmake process hence their usage limitation

PR #208, opened 2019-05-07
[IDL] enable docstrings only in Python 3

nim65s:
as attribute '__doc__' of 'type' objects is not writable in Python 2


jcarpent:
cc @ylabbe

jmirabel:
This looks good to me.

PR #207, opened 2019-05-04
Fixes #206

mkatliar:

Issue #206, opened 2019-05-03
boost_python library could not be found on Ubuntu 19.04


mkatliar:
I get the following error from CMake trying to find boost python libraries:

```
CMake Warning at /usr/share/cmake-3.13/Modules/FindBoost.cmake:1775 (message):
  No header defined for python-py37; skipping header check
Call Stack (most recent call first):
  cmake/boost.cmake:104 (FIND_PACKAGE)
  CMakeLists.txt:169 (SEARCH_FOR_BOOST)


CMake Error at /usr/share/cmake-3.13/Modules/FindBoost.cmake:2100 (message):
  Unable to find the requested Boost libraries.

  Boost version: 1.67.0

  Boost include path: /usr/include

  Could not find the following Boost libraries:

          boost_python-py37

  Some (but not all) of the required Boost libraries were found.  You may
  need to install these additional Boost libraries.  Alternatively, set
  BOOST_LIBRARYDIR to the directory containing Boost libraries or BOOST_ROOT
  to the location of Boost.
Call Stack (most recent call first):
  cmake/boost.cmake:104 (FIND_PACKAGE)
  CMakeLists.txt:169 (SEARCH_FOR_BOOST)


CMake Error at cmake/boost.cmake:107 (MESSAGE):
  Failed to detect Boost with the following components:

Call Stack (most recent call first):
  CMakeLists.txt:169 (SEARCH_FOR_BOOST)
```

The library name on my system is `boost_python37`, not `boost_python-py37`. 

Changing the lines
```cmake
IF(APPLE OR NETBSD_FOUND OR ARCHLINUX_FOUND)
          SET(BOOST_PYTHON_COMPONENT "python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
        ELSEIF(DEBIAN_FOUND)
          SET(BOOST_PYTHON_COMPONENT "python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
        ELSE(APPLE OR NETBSD_FOUND OR ARCHLINUX_FOUND)
          SET(BOOST_PYTHON_COMPONENT "python${PYTHON_VERSION_MAJOR}")
        ENDIF(APPLE OR NETBSD_FOUND OR ARCHLINUX_FOUND)
```
to
```cmake
IF(APPLE OR NETBSD_FOUND OR ARCHLINUX_FOUND OR DEBIAN_FOUND)
[...]

PR #213, opened 2019-07-11
Boost-python: change rules to search for it

jcarpent:
Related to #207 and #206.


jcarpent:
Yes, indeed. But he is not available.
I will merge it and provide some fixes if needed.

PR #214, opened 2019-07-11
python: use default architecture policy to build PYTHON_SITELIB

jcarpent:
@jmirabel @wxmerkt I made a proposal to be platform dependent in terms of computing the PYTHON_SITELIB variable.
I use Python to retrieve the correct destination. 
Can you provide some feedbacks on it?


python.cmake
```diff
@@ -77,25 +77,20 @@ GET_FILENAME_COMPONENT(PYTHON_LIBRARY_DIRS "${PYTHON_LIBRARIES}" PATH)
...
 OPTION(PYTHON_DEB_LAYOUT "Enable Debian-style Python package layout" OFF)
 
-IF (PYTHON_DEB_LAYOUT)
-  SET(PYTHON_PACKAGES_DIR dist-packages)
-ENDIF (PYTHON_DEB_LAYOUT)
-
 EXECUTE_PROCESS(
   COMMAND "${PYTHON_EXECUTABLE}" "-c"
-  "import sys, os; print(os.sep.join(['lib', 'python' + sys.version[:3], '${PYTHON_PACKAGES_DIR}']))"
+  "from distutils import sysconfig; print(sysconfig.get_python_lib(prefix='', plat_specific=True))"
```
wxmerkt: `plat_specific=True` means e.g. `lib64` vs `lib` (`site-packages` vs `dist-packages` is OS-specific). Is that intended and do you need that platform-specific for your use cases - it will change existing lib to lib64 etc.? [just tested on Ubuntu 18.04 and RHEL]

python.cmake
```diff
@@ -77,25 +77,20 @@ GET_FILENAME_COMPONENT(PYTHON_LIBRARY_DIRS "${PYTHON_LIBRARIES}" PATH)
...
   OUTPUT_VARIABLE PYTHON_SITELIB
+  OUTPUT_STRIP_TRAILING_WHITESPACE
   ERROR_QUIET)
 
-# Remove final \n of the variable PYTHON_SITELIB
-STRING(REPLACE "\n" "" PYTHON_SITELIB "${PYTHON_SITELIB}")
-NORMALIZE_PATH(PYTHON_SITELIB)
+# Keep compatility with former jrl-cmake-modules versions
+IF(PYTHON_DEB_LAYOUT)
```
wxmerkt: With the above command, is this option still required or doing anything? On Debian, the above command should give you `dist-packages` already (or is this for cross-compiling?)

jcarpent: I know that people from JRL are still dependent on this variable. I prefer to keep compatibility with them. But I do agree, this is no more needed.

python.cmake
jcarpent: You're right. I will change it. Thanks for this quick feedback.

jcarpent:
@wxmerkt I've pushed the changes.

python.cmake
jcarpent: Done!

wxmerkt: Sorry, I am not seeing the change reflected here - did you already push?

jcarpent: Sorry, I indeed forgot to push. Thanks :)

jcarpent:
I will merge it soon. Thanks @wxmerkt for your reviews.

wxmerkt:
Thank you very much for this fix, I just tested on Ubuntu 18.04 and it works perfectly :-)

nim65s:
FTR, this breaks robotpkg settings: `lib/python3.6/site-packages` has been turned into `lib/python3/dist-packages`.

Having `PYTHON_PACKAGES_DIR` allows to revert `dist` to `site`, but we might need another option to get the python minor version back.

PR #218, opened 2019-07-16
fix some issue related to recent versions of CMake

jcarpent:


base.cmake
```diff
@@ -175,7 +167,11 @@ MACRO(SETUP_PROJECT)
   ENDFOREACH(VARIABLE)
 
   # Define project name.
-  PROJECT(${PROJECT_NAME} CXX)
+  IF(${CMAKE_VERSION} VERSION_GREATER 3.9.0)
+    PROJECT(${PROJECT_NAME} LANGUAGES CXX DESCRIPTION ${PROJECT_DESCRIPTION})
+  ELSE(${CMAKE_VERSION} VERSION_GREATER 3.9.0)
+    PROJECT(${PROJECT_NAME} CXX)
+  ENDIF(${CMAKE_VERSION} VERSION_GREATER 3.9.0)
```
jmirabel: For reference, there is a warning with recent versions of CMake about the PROJECT_VERSION variable. This variable should be set via the argument VERSION of function PROJECT.

PR #222, opened 2019-07-16
set DISABLE_TESTS to OFF when BUILD_TESTING is ON

jcarpent:


test.cmake
```diff
@@ -35,6 +35,11 @@ IF(NOT DEFINED DISABLE_TESTS)
   SET(DISABLE_TESTS OFF)
 ENDIF(NOT DEFINED DISABLE_TESTS)
 
+# Active test suites if BUILD_TESTING is ON
+IF(BUILD_TESTING)
+  SET(DISABLE_TESTS OFF)
+ENDIF(BUILD_TESTING)
```
jmirabel: `BUILD_TESTING` is a CMake variable, right ?

I would either do:
```cmake
IF(NOT DEFINED DISABLE_TESTS)
  IF(${BUILD_TESTING})
    SET(DISABLE_TESTS OFF)
  ELSE()
    SET(DISABLE_TESTS ON)
  ENDIF()
ENDIF(NOT DEFINED DISABLE_TESTS)
```
or
```cmake
IF(DEFINED DISABLE_TESTS)
  MESSAGE(AUTHOR_WARNING "DISABLE_TESTS is deprecated. Use BUILD_TESTING instead.")
  IF(${DISABLE_TESTS})
    SET(BUILD_TESTING OFF)
  ELSE()
    SET(BUILD_TESTING ON)
  ENDIF()
ENDIF(DEFINED DISABLE_TESTS)
```

jcarpent: I changed the files to reflect your proposition.
Thanks for your review.

jcarpent:
This PR changes the behavior by using as most as possible the CMake BUILD_TESTING variable.

PR #223, opened 2019-07-17
Fix Windows\MSVC build issues.

seanyen:
This pull request is to address build issues when I am building [`eigenpy`](https://github.com/stack-of-tasks/eigenpy) on Windows with MSVC and [`ROS catkin environment`](https://aka.ms/ros).

* `Boost_<C>_LIBRARY` could include `optmized\debug` keywords so I chose to use per-configuration variables `Boost_<C>_LIBRARY_DEBUG` or `Boost_<C>_LIBRARY_RELEASE` to accommodate it.
* Normalize `${NUMPY_INCLUDE_DIRS}` so it won't contain backslash characters which caused CMake `Invalid escape sequence` error.
* Avoid to double define `ADD_CUSTOM_TARGET(build_tests)` which caused CMake error.
* Fixed incompatible defines for `pkgconfig-lite` in `shared-library.cmake` on Windows build.


shared-library.cmake
```diff
@@ -16,10 +16,11 @@
 # Shared library related constants
 # (used for pkg-config file generation).
 # FIXME: can't we get these information from CMake directly?
+SET(LIBDIR_KW "-L")
+SET(LIBINCL_KW "-l")
+SET(LIBINCL_ABSKW "")
+SET(LIB_EXT "")
```
jmirabel: @jcarpent Those default values are fine for Mac ?

shared-library.cmake
```diff
@@ -16,10 +16,11 @@
...
 IF(WIN32)
-  SET(LIBDIR_KW "/LIBPATH:")
-  SET(LIBINCL_KW "")
-  SET(LIBINCL_ABSKW "")
   SET(LIB_EXT ".lib")
 ELSEIF(UNIX)
   SET(LIBDIR_KW "-L")
```
jmirabel: This line and the one below can be removed.

shared-library.cmake
jcarpent: I will try them indeed.

jcarpent: Seems fine.

shared-library.cmake
jcarpent: I will do that now.

seanyen:
@jcarpent Yes, that would be lovely! I verified and it builds on `devel` branch up to this commit: https://github.com/stack-of-tasks/eigenpy/commit/48ff1fd19ecff61f4dbffc1fe5bda4e406369e56

PR #226, opened 2019-07-28
Properly appending the list of python paths

cmastalli:
This fixes an important bug for Python-base unit-test code.

The previous method didn't add the `:` symbol needed between different path.


cmastalli:
@nim65s This solves the issue that we discussed few weeks ago.

PR #230, opened 2019-08-08
[HPP] Add macro to homogenize the finding of Python.

jmirabel:


hpp/python.cmake
```diff
@@ -0,0 +1,27 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+IF(NOT COMMAND FINDPYTHON)
+  include(cmake/python.cmake)
+ENDIF()
+
+SET(PYTHON_DEFAULT_VERSION 2.7)
```
jcarpent: You will need to change it soon, as Python 2.x won't be maintained anymore after January, 1st 2020.

jmirabel: Yes, I know. I'm almost finished porting to Python 3 so Python 2 is still the default.

PR #232, opened 2019-08-27
Revert "[HPP] Add macro to homogenize the finding of Python."

nim65s:
Hi,

Before this PR, it was already possible to choose python version with:
`cmake -DPYTHON_EXECUTABLE=/usr/bin/python2 ..` or
`cmake -DPYTHON_EXECUTABLE=/usr/bin/python3 ..` (or any other python executable)
Which means that by default, the python version will be the version of the `python` executable which is the first in the `$PATH`, so I don't see the point in introducing this `$PYTHON_DESIRED_VERSION`


nim65s:
ref. https://github.com/jrl-umi3218/jrl-cmakemodules/pull/167#issuecomment-433921813

jmirabel:
I didn't know about that. I am fine with PYTHON_EXECUTABLE.

nim65s:
My bad for not reacting earlier

PR #233, opened 2019-09-03
add  INSTALL_GENERATED_HEADERS & INSTALL_PKG_CONFIG_FILE options

nim65s:
Hi,

For a package which has python bindings, it is convenient to build on
the one hand only the main binary package, and on the other hand several
binary packages with the bindings for different python versions.

For this other packages to be able to co-exist on a system (and ease
transition from Python 2 to Python 3), there must not be any path
conflict, ie. everything must be either in a PYTHON_SITELIB or with a
SOABI depending on the python version.

The automatically generated
- lib/pkgconfig/<project>.pc
- include/<project>/config.hh
- include/<project>/deprecated.hh
- include/<project>/warning.hh
files are preventing this to work.

This PR adds a BINDINGS_ONLY option allows to avoid the installation of
those files.


jmirabel:
I have a few remarks:
1. do you need to generate the headers ? Otherwise, you can also deactivate their generation.
2. I would split the variable into PROJECT_INSTALL_HEADERS and PROJECT_INSTALL_PKG_CONFIG_FILE (or PROJECT_GENERATE_PKG_CONFIG_FILE). BINDINGS_ONLY is not very helpful in understanding what it does...

nim65s:
Agreed for both

nim65s:
I named those `INSTALL_GENERATED_HEADERS` & `INSTALL_PKG_CONFIG_FILE`, to follow the existing `INSTALL_DOCUMENTATION`

nim65s:
after reflexion, I found 
https://github.com/loco-3d/sot-talos-balance/blob/6b0d78ccbf3f1e69a20fee85eb3110d028d18a05/CMakeLists.txt#L61

So this might still be useful to have an option that would disable the standard headers, and the standard .pc, and any other package-depend file.
And maybe documentation too, as it is also installed out of PYTHON_SITELIB.

Should I add an `OPTION(INSTALL__PYTHON_INTERFACE_ONLY)`, which would set defaults for `INSTALL_DOCUMENTATION`, `INSTALL_GENERATED_HEADERS` and `INSTALL_PKG_CONFIG_FILE` ?

jmirabel:
I don't think the variable name should contain anything related to python. You may add something like `dont_install_anything_by_default`.

nim65s:
That's right for the headers and the .pc.

Then, I could easily keep the definition of `INSTALL_PYTHON_INTERFACE_ONLY` in packages, and turn off the 3 others there.

nim65s:
@jcarpent this has already been done,  following @jmirabel ideas

jcarpent:
Sounds good to me then. Sorry for the noise.

PR #236, opened 2019-09-17
test: pass the make options to cmake --build

jcarpent:


jcarpent:
> That supposes that you use CMake and Make. I guess this supposition is fine.

Yes, indeed. We can see later to handle other builder.

PR #238, opened 2019-09-20
Allowed to build Python scripts with same names but differents scopes

cmastalli:


cmastalli:
This PR uses the scope of the current Python script to rename the target file. This is important if you have two Python scripts with the same name but living in different folders.

cmastalli:
@jcarpent or @jmirabel could you review this PR? It just a tiny change, it shouldn't take more than couple of minutes

PR #239, opened 2019-09-25
PYTHON_INSTALL install only the source, fix #136

### LINKED ISSUE #136: [CMake] Installing .pyc files
Hi,

In https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/python.cmake#L274 we try to install .pyc files.

First of all, I don't understand why.
Does someone have an explanation ?

Then, in python 3.2+, [PEP 3147](https://www.python.org/dev/peps/pep-3147/) states that compiled files are no more `$filename.pyc`, but `__pycache__/$filename.<magic>.pyc` with `<magic>` something like `cpython-35`, so cmake will not find `$filename.pyc`

This issue follows https://github.com/stack-of-tasks/pinocchio/issues/464

nim65s:
and deprecate PYTHON_INSTALL_BUILD

### DIFF
--- a/python.cmake
+++ b/python.cmake
@@ -308,7 +308,7 @@ ENDMACRO()
 #
 MACRO(PYTHON_INSTALL_BUILD MODULE FILE DEST)
 
-  MESSAGE(AUTHOR_WARNING "PYTHON_INSTALL_BUILD is deprecated and will be removed")
+  MESSAGE(AUTHOR_WARNING "PYTHON_INSTALL_BUILD is deprecated and will be removed in the future")
   MESSAGE(AUTHOR_WARNING "Please use PYTHON_INSTALL_ON_SITE")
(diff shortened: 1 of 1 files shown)


python.cmake
```diff
@@ -309,6 +308,10 @@ ENDMACRO()
 #
 MACRO(PYTHON_INSTALL_BUILD MODULE FILE DEST)
 
+    MESSAGE(WARNING "PYTHON_INSTALL_BUILD is deprecated and will be removed")
+    MESSAGE(WARNING "Please use PYTHON_INSTALL_ON_SITE")
+    MESSAGE(WARNING "ref https://github.com/jrl-umi3218/jrl-cmakemodules/issues/136")
```
jmirabel: Replace WARNING by AUTHOR_WARNING here.

nim65s: done :)

PR #240, opened 2019-09-30
ADD_PYTHON_UNIT_TEST: allow multiple paths in argn

nim65s:
Hi,

Multiple paths can be set in PYTHONPATH, but they have to be joined by `os.pathsep,`, which is `:` on unix, and `;` on windows ([ref](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH)).

This PR fix the concatenation with any pre-existing environment variables, and allows to set several folders (relative to the project build tree) inside this PYTHONPATH.

eg. if you build a module in `src/` and a test module in `tests/`, you can import them both in a test with:
`ADD_PYTHON_UNIT_TEST("test-python" "tests/main.py" "src" "tests")`


test.cmake
```diff
@@ -98,16 +98,23 @@ MACRO(ADD_PYTHON_UNIT_TEST NAME SOURCE)
...
+  ENDFOREACH(MODULE_PATH IN LISTS MODULES)
+
   IF(DEFINED ENV{PYTHONPATH})
-    SET(PYTHONPATH "${PYTHONPATH}:$ENV{PYTHONPATH}")
+    LIST(APPEND PYTHONPATH "$ENV{PYTHONPATH}")
   ENDIF(DEFINED ENV{PYTHONPATH})
 
-  SET(ENV_VARIABLES "PYTHONPATH=${PYTHONPATH}")
+  # get path separator to join those paths
+  EXECUTE_PROCESS(COMMAND
```
jcarpent: Is this line has to rely on Python?

nim65s: This is is the more portable way to do that, and Python is needed anyway at this point: `PYTHON_EXECUTABLE` is used in the first line of this macro.

We could hardcode the results, but I don't see the benefits.

Issue #188, opened 2019-02-18
Update to more recent CMake version


jmirabel:
This is not of high priority but, for some of them, we regularly get warning messages. If relevant, I will try to update the list as I discover changes.
- [x] the project version should be passed to macro `project` (from 3.0 I think).
- [x] CMake targets of the form `boost::python` rather than providing the path.

jcarpent:
I get the first point, but not the second point. Can you explain a bit further?

jmirabel:
> I get the first point, but not the second point. Can you explain a bit further?

I can't since I don't know the details. I know they are called imported targets.

jmirabel:
Use [GENERATE_EXPORT_HEADER](https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html) to generate `config.hh`.

jcarpent:
Since CMake 3.15, the call to the PROJECT should be done at the first lines of the root CMakeLists.txt.
See https://cmake.org/cmake/help/v3.15/command/project.html#code-injection for further explanations.

jmirabel:
As I understand the doc, it is forbidden to call PROJECT from an included file. I think calling PROJECT inside a macro called from the root CMakeLists.txt is fine as macro are not functions.

jcarpent:
This is currently warning on my Mac currently:
```
CMake Warning (dev) in CMakeLists.txt:
  No project() command is present.  The top-level CMakeLists.txt file must
  contain a literal, direct call to the project() command.  Add a line of
  code such as

    project(ProjectName)

  near the top of the file, but after cmake_minimum_required().

  CMake is pretending there is a "project(Project)" command on the first
  line.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

jmirabel:
So we are doomed...

PR #242, opened 2019-10-08
Make it possible to call project() in the root CMakeLists.txt

gergondet:
This is related to #188 

This deprecates the use of setup_project() in favor of:

```
set(PROJECT_NAME "MyProject")
set(PROJECT_DESCRIPTION "...")
set(PROJECT_URL "...")
include(cmake/base.cmake)
project(${PROJECT_NAME} CXX)
```

which is going to be CMake friendly from 3.15 onwards

Also, `setup_project_finalize()` and `setup_project_package_finalize()` are now automatically called at the end of the configuration phase thanks to a nice trick I found on SO. The later is only called is `PROJECT_USE_CMAKE_EXPORT` is set to `TRUE`

To do so, including cmake/base.cmake will have the following effect:
- setup a script to call after project() is called
- compute PROJECT_VERSION and restore its value after project() is called (we cannot use the VERSION argument of project() because we compute incompatible version numbers)
- setup a hook to call setup_project_finalize() and optionally setup_project_package_finalize()

I think this is a breaking change for most projects because we currently do:

```
include(cmake/base.cmake)
# Include other things
set(PROJECT_NAME "...")
# Set other PROJECT_ variables
setup_project()
```

A quick fix in those projects is to set the variables before including `cmake/base.cmake` and optionally replace the `setup_project()` call. I could make a backward compatible change but I feel it's more trouble than it's worth. CMake will fail very early and very explicitly but I could make some explicit suggestion that you are upgrading from "old-ways".

Note: I realized (late because I was working from a slightly outdated master) that there was some recent changes to make calls to `project()` with `LANGUAGES` and `VERSION` arguments depending on CMake version. Since I don't get any warning in CMake 3.15 and CMake 3.5 for doing `project()` calls the old fashion way I'm assuming this is not really necessary. I just make sure to set `PROJECT_VERSION` back to what we want it after the `project()` call.


vsamy:
Nice work!

jmirabel:
That is a very nice work.

> I could make a backward compatible change but I feel it's more trouble than it's worth.

I agree.

> CMake will fail very early and very explicitly but I could make some explicit suggestion that you are upgrading from "old-ways".

I do not understand what you meant. If possible, I would be in favor of adding a CMake developer warning, with a URL pointing to some page of the doc that explains what action should be taken to update the CMake module. Is this possible ? Would you mind doing this ?

gergondet:
> I do not understand what you meant. If possible, I would be in favor of adding a CMake developer warning, with a URL pointing to some page of the doc that explains what action should be taken to update the CMake module. Is this possible ? Would you mind doing this ?

What will happen is that when you include `cmake/base.cmake` before setting `PROJECT_*` variables, it will fail because the variables are not defined (i.e. very explicitly) and since we usually have `include(cmake/base.cmake)` within the first few lines, it will fail very early.

The message you get right now is:
```
MESSAGE(FATAL_ERROR "Required variable ``${VARIABLE}'' has not been defined.")
```

I'm suggesting is to change it to:
```
MESSAGE(AUTHOR_WARNING "Required variable ``${VARIABLE}'' has not been defined, perhaps you are including cmake/base.cmake too early")
MESSAGE(FATAL_ERROR "Required variable ``${VARIABLE}'' has not been defined.")
```

jmirabel:
A link to https://jrl-cmakemodules.readthedocs.io/en/master/pages/base.html#minimal-working-example might be useful in the warning.

What happens for CMake < 3.15 ? Is the CMake minimum version really 2.8 ?
My only concern is to have something which works for CMake < 3.15 and CMake > 3.15.

I am open to increasing the minimum required CMake version if necessary.

gergondet:
> A link to https://jrl-cmakemodules.readthedocs.io/en/master/pages/base.html#minimal-working-example might be useful in the warning.

I agree. I have added an `AUTHOR_WARNING` message when the required variables are not defined

> What happens for CMake < 3.15 ? Is the CMake minimum version really 2.8 ?
My only concern is to have something which works for CMake < 3.15 and CMake > 3.15.

I've tested it with CMake 3.5 (on Ubuntu 16.04) and as far as I can tell, it works.

From 2.8.12 doc it should also work nicely since the `CMAKE_PROJECT_${PROJECT_NAME}_INCLUDE` trick that it relies on is documented this far back.

I'm not sure about the post-configuration hook for CMake 2.8 but I'm guessing it should work as well (`variable_watch` is an old command so that's ok, the only uncertainty is whether CMake has the same behavior but the SO answer I relied upon for the trick is dated 2013 so that's around CMake 2.8 ^^)

jmirabel:
Perfect. Sounds all good to me.

jmirabel:
We should tag the master branch just before this PR is merged. It makes it easy to go back to the latest backward compatible version if needed.

gergondet:
FYI, in a project that mixes C and C++ codes, I had to change `project(${PROJECT_NAME} CXX)` to `project(${PROJECT_NAME} C CXX)`. This was not necessary before in the same project and I don't have time to investigate too deeply a minor annoyance ^^

> We should tag the master branch just before this PR is merged. It makes it easy to go back to the latest backward compatible version if needed.

Sure. Any suggestion on the tag name? `before_setup_project_deprecation`?

I'll tag and merge once we agree.

jmirabel:
> `before_setup_project_deprecation`?

Thats fine for me.

gergondet:
> This was not necessary before in the same project and I don't have time to investigate too deeply a minor annoyance

Just a guess... Not having `setup_project()` in the main CMakeLists.txt meant that CMake was defaulting to the C/CXX language



>>    `before_setup_project_deprecation`?
>
> Thats fine for me.

Done. I will merge this now, thanks for the feedback.

PR #244, opened 2019-10-10
os: split in a separate file the definition of OS checker

jcarpent:
I need to define first the CHECK_{DEBIAN,NETBSD,etc} first before including `INCLUDE(${CMAKE_CURRENT_LIST_DIR}/GNUInstallDirs.cmake)`


nim65s:
cc @olivier-stasse

jcarpent:
@nim65s @nim65s Did you encounter the same issue?

nim65s:
I was able to make a release without issue, but at the same time @olivier-stasse got an issue… So I wanted to investigate on the differences, but if you just fix it that's fine for me :)

jcarpent:
It seems working on my side. I will merge it to move forward.

PR #246, opened 2019-10-11
[tests] Correctly set BUILD_TESTING when DISABLE_TESTS is passed

gergondet:
This PR correctly sets `BUILD_TESTING` when only `DISABLE_TESTS` is passed to cmake

In the current version, ctest is included after we set the `BUILD_TESTING` variable according to the `DISABLE_TESTS` variable and this ends up not honoring `DISABLE_TESTS` until CMake is run a second time


gergondet:
@jmirabel FYI, documentation hook is working again

PR #250, opened 2019-10-22
Fix CMake errors on Windows

seanyen:
On Windows build, `<substring>` without quoted will be ignored by `CMake`, and results in the following errors:

```
CMake Error at cmake/test.cmake:123 (STRING):
  STRING sub-command REPLACE requires at least four arguments.
  unittest/CMakeLists.txt:48 (ADD_PYTHON_UNIT_TEST)
```


gergondet:
Thanks for the fix!

PR #251, opened 2019-10-22
Make FINDPYTHON more portable for Windows and Linux

seanyen:
This pull request is attempting to make `FINDPYTHON` more portable for both Windows and Linux.

* In favor of `EXT_SUFFIX` over `SOABI`, because only `EXT_SUFFIX` exists for both Windows and Linux for Python3.
* If `EXT_SUFFIX` doesn't exist (most likely it is Python2), then fallback to `.pyd` for Windows and `.so` for the rest (most likely it is Linux or Mac OSX).


seanyen:
This change would require additional changes on `Eigenpy` too. I will be sending it out shortly for review.

jmirabel:
> This change would require additional changes on `Eigenpy` too. I will be sending it out shortly for review.

This may impact more than eigenpy. Could you set PYTHON_SOABI to PYTHON_EXT_SUFFIX, with a comment saying this is for backward compatibility and that it should be removed later  ?

jmirabel:
Moreover, could you add something like this after the documentation of PYTHON_SITELIB at the beginning of the file:
```
#.rst:
# .. variable:: PYTHON_EXT_SUFFIX
#
#  Portable suffix of C++ Python modules.
```

seanyen:
> This may impact more than eigenpy. Could you set PYTHON_SOABI to PYTHON_EXT_SUFFIX, with a comment saying this is for backward compatibility and that it should be removed later ?

Sure thing! Since `PYTHON_SOABI` and `PYTHON_EXT_SUFFIX` has different meanings, so I brought back the original `PYTHON_SOABI` but skip it on Windows, because it causes `CMake` error anyway.

Also I added the argument comments as extra documentation.

gergondet:
Merging.

@seanyen you'll need to update the submodule on stack-of-tasks/EigenPy#97 to get the changes

PR #258, opened 2019-10-30
version: read version from package.xml if needed

jcarpent:


jcarpent:
@wxmerkt This might be of your interest.

jcarpent:
I will merge it as it is required by Pinocchio on ROS build systems.

wxmerkt:
Thank you! :-)

jmirabel:
I agree this solves a problem but it makes the code more complex. This will be troublesome for someone in the future.

I would put the code into a macro to be called in a project if the developer thinks it is relevant.

jcarpent:
The issue is coming from the git history which is not loaded.
As @wxmerkt had issue with ROS building, I suggested to read the version from the package.xml. So, it solved the issue of @wxmerkt for the moment. 

The main problem is coming from the downloader which provides a fix and limited depth to the git tree (without tags).

jmirabel:
I think it is the wrong way of fixing this issue.

jmirabel:
Having a different behavior based on whether you cloned the repo with `git clone` or with `git clone --depth 3` isn't satisfactory.

jcarpent:
But I think this should be good to take the version from package.xml, this is not incorrect.

jmirabel:
Why not always reading from package.xml and not using `git describe` then ?

jcarpent:
Because not all the projects have their own pakcage.xml.

jmirabel:
I know and I do not want anyone to have one.

What I prefer is that :
1. the version is always read from the same place independently of the compilation environment.
2. we give preference to version number entered by hand over version number guessed with any method, including `git describe`.

If you enter a version number by hand in package.xml. Read it, always.

I don't think it is a good idea to do something like: *if you have a file called `package.xml`, read it*. i think it is better to provide a macro that:
- is used manually, project by project,
- makes it clear that package.xml is in a given format (i.e. the macro name could be `COMPUTE_VERSION_FROM_ROS_PACKAGE_FILE`).

jcarpent:
One solution would be to select different ways (ordered) to retrieve the version.
- .version
- git
- others ...

wxmerkt:
What happens if someone takes a tar/zip ball?

The current order already does git describe, if fails check .version, if fails fail. This adds one more if-else. It is a work-around for trying to rely on git describe which should not be taken as "there always will be tags in a repository"

olivier-stasse:
I would prefer that we have a consistent policy on the versioning.
Incrementing the source of version is likely to increase the mistake.

I run into many problems when .version where present in some repositories, specifically when doing integration tests.

For me the version should be register in the history of the repo, whatever versioning system is used.
Having static file providing information have only one interest: they do not fail when used by build system.
Thus maybe we should have macros checking the consistency of the version with git for files specifying the version when performing release.

Issue #253, opened 2019-10-24
git describe failing causes project to not compile


VladimirIvan:
When git describe fails, code stops compiling as the version numbers cannot be filled in. This is the case e.g. on a new repository with branches but no tags. 

cc @wxmerkt

jcarpent:
Small question, what is the output of `git tag`? 
In fact, the module is looking for the current version of the project either in the file named `.version`in the root directory of the project or inside the git tree (a tag to be more precise).

jmirabel:
I think I already encountered a similar issue. I solved it be setting PROJECT_VERSION in the CMakeLists.txt.
IMHO: The git describe message is only a warning and the issue you encounter comes either from #242 or #221

jcarpent:
So if you put the tags on the fork, it should work. But maybe, you do not want to?

wxmerkt:
Yes, putting the tags makes it work. It would be good if not having tags or requiring a repository was handled gracefully. For instance, some workflows (including the buildfarm) involve downloading the tar.gz archive which by nature isn't a git repository.

jcarpent:
Sure. Normally, the package generated by the jrl cmake module is taking care for the generation of the `.version`file to get the value.

jmirabel:
> Yes, putting the tags makes it work. It would be good if not having tags or requiring a repository was handled gracefully. For instance, some workflows (including the buildfarm) involve downloading the tar.gz archive which by nature isn't a git repository.

That archive should contain a `.version` file at the root.

jmirabel:
I don't think #258 changes anything to this.

jcarpent:
This is is not an archive, but an issue of git tree. See #258 for more details.

wxmerkt:
> That archive should contain a .version file at the root.

I see neither pinocchio nor eigenpy including `.version` files.

jcarpent:
It is present in the tar.gz of the releaes ;à

wxmerkt:
I see. Is it manually added? People who download master/devel as a zip-ball per github wont get it though as it's not checked in, correct?

jcarpent:
It is added by the command 'make release VERSION=x.y.z` of the submodule. It will create a tar.gz with all the submodules of git integrated, as by default the packager of GitHub is doing this job.

olivier-stasse:
[...]

PR #264, opened 2019-11-04
Fix shallow git repository issues

jcarpent:
related to #253 or #258.

### DIFF
--- a/version.cmake
+++ b/version.cmake
@@ -53,9 +53,6 @@
 #    - ``0.5-2-034f`` if there is no uncommitted changes,
 #    - ``0.5-2-034f-dirty`` if there is some uncommitted changes.
 #     
-#    If the current repository is a shallow copy, then the function git fecth --unshallow is called to
-#    allow the computation of version from a git tag.  
-#
 #  - the software comes with a package.xml file at the root of the project (for ROS build essentially)
 #    then the module extracts the version number which is declared inside between the tag <version>x.y.z<\version>
 #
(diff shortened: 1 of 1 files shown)


wxmerkt:
This may not work on the buildfarm as the source directory is likely to be mounted read-only.

jcarpent:
Are you sure of that?

jmirabel:
It doesn't really change my comment https://github.com/jrl-umi3218/jrl-cmakemodules/pull/258#issuecomment-549303507

I would revert the changes in VERSION_COMPUTE and add another macro that compute the version number from package.xml. I don't really like to automatically guess everything. It is too brittle in this context.

jcarpent:
@jmirabel I do understand your point. I'm first trying to fix what is missing to make git describe more robust, mostly when the git repo is not a deep copy of the original one.

wxmerkt:
The CMake clear way would be to define the version in the CMake, or perhaps the `.version`. Reading it from git describe was kind of the first "brittle" sin which the earlier PR expanded on (if you want to take this viewpoint - the other is, it adds another layer of graceful degradation/robustness before fully failing). I think Justin's fix of "if-git-describe-fails-see-if-package-xml-exists" is a nice graceful way before failing - especially since there is no CMake-error-message/configuration error but a compile (!) error (variable not defined). 

And yes, we should not try to consider the source directory to be writable. It is good practice to consider the source tree to be read-only.

jmirabel:
> The CMake clear way would be to define the version in the CMake

This can still be done project-wise.

jcarpent:
The idea of externalizing the version definition is to allow an easy way to automatize the creation of release with `make release` for instance.
But, we can of course provide a script to perform as well on the CMakeLists with the creation of a commit.

wxmerkt:
Ok, I wasn't aware of the `make release` call. Yet, to come back to this PR, it won't solve the issue encountered - we cannot assume the source tree to be writable. The graceful-failback you've previously added does work (and in my opinion should not be deactivated).

jcarpent:
Following your feedback, I will just raise a warning to detect if the repository is shallow.

jcarpent:
@jmirabel I have just updated the PR.

jmirabel:
I won't block you on this but I still think we have a very complicated way to get the version number. Adding package.xml as an automatic fallback solves one issue and adds confusion.

jcarpent:
I will deal with this concern in an other PR.

PR #270, opened 2019-11-07
version: robustify the handling of UNKNOWN tag

jcarpent:


gabrielebndn:
I hink this is not 100% robust. If version computation succeeds, then `PROJECT_VERSION_[...]_CONFIG` is an integer literal, whereas if it fails then it is a string literal. This is not 100% compatible and it may still cause compilation failures and/or unexpected bugs.
I think it would be better to just set them to zero.

jcarpent:
The goal is then to deal with that inside your code. With something like:
```
#if PINOCCHIO_MAJOR_VERSION == PINOCCHIO_VERSION_UNKNOWN_TAG
// do what you want
#endif
```

gabrielebndn:
I do not think you can easily perform string comparison in preprocessor directives.
It was probably easier with just the empty `UNKNOWN` macro
https://stackoverflow.com/questions/2335888/how-to-compare-strings-in-c-conditional-preprocessor-directives

gabrielebndn:
I confirm that, before this change, `#if PINOCCHIO_MAJOR_VERSION == UNKNOWN` (or  `#if PINOCCHIO_MAJOR_VERSION != UNKNOWN`) was enough to check whether the version was successfully detected or not.
On the other hand, this change does not solve compilation errors in place where integers are required and checking a string literal on the C preprocessor is much more complicated.
Overall, I believe this change to be harmful. It should either be reverted, or the version numbers should be 0.0.0 if not properly detected. Also, macro `@LIBRARY_NAME@_VERSION` should be handled consistently.

config.hh.cmake
```diff
@@ -26,10 +26,11 @@
 # define @LIBRARY_NAME@_CONFIG_HH
 
 // Package version (header).
+# define @LIBRARY_NAME@_VERSION_UNKNOWN_TAG // Used to mention that the current version is unknown.
```
jmirabel: @jcarpent thank you for the work you do for the community. However, could you please be less quick in merging such changes ?
Those change should be discussed. I strongly disapprove the idea of having this into the C++ code. Version computation **must** be handled at configure time.
If git does not fulfill all the requirement to compute the current version, then we should think of another method.

jcarpent: This fix allows us to provide Pinocchio on ROS for instance. We had serious limitations due to ROS build, and in this way, we were able to handle it. The problem of computing the version is that currently, we do not have a common consensus in where to store it, knowing that the file `.version` is only provided in Releases, and it is not inside the package.

Maybe for each package, we should have something like package.xml which provided useful information like main contributors, website, etc.

I'm not against changing the strategy in future, but I would like the packages remain available for the community.

jmirabel: > The problem of computing the version is that currently, we do not have a common consensus in where to store it

and we should solve that before doing so many changes. If you want a temporary fix in Pinocchio before this consensus is found, just set the version explicitly in the CMakeLists.

jcarpent: It will appear in other places that have copied pasted the code from Pinocchio. So this is a temporay fix I agree, but will avoid other burden outside ;)

jmirabel: A very simple temporary fix would have been `SET(PROJECT_VERSION ...)` in Pinocchio and EigenPy. There was no need to dirty the repo.

gabrielebndn:
Some time ago (about an year), the version numbers where hard-coded inside the `config` header. As a result, we regularly forgot updating them when making a release (see https://github.com/stack-of-tasks/pinocchio/issues/551). Computing them with git from the tags (https://github.com/jrl-umi3218/jrl-cmakemodules/pull/162) has solved this problem for good. If we go back to hard-coding the version somewhere, we will again have the same problem as before.

gabrielebndn:
It may sound a bit crazy, but here is a proposition:
1. Have a file `package.xml` (or whatever) at the root of our packages.
2. `make release` takes care of modifying this file and committing the change before doing the release.
3. For a normal build, `package.xml` is read to retrieve the version.

Is it doable? Would it cause more problems than it solves?

jmirabel:
I am fine with #162. CMake computes a version number and configures a header. This is clean.

Talking about pinocchio, you now have the version number in the tags and in the package.xml so it is not better.

The question is about how our CMake module computes the version number. We use `git describe` for this but there are cases where it fails (and we should not assume this will always work). So we have several options:
- add a failback (like package.xml) but this means the version number will be into 2 places.
- always read it from package.xml no matter what the tag is.
- hard code it in the CMakeLists.
- any options I may not have in mind.

To me, the best is to provide a macro for each solution and let the project decide what fits their needs.

What I am asking for is a clear discussion on that and not a sprint where developers are involved only if they answer within the next 2 hours a PR is opened.

gabrielebndn:
I agree it should be discussed. I think a cmake option choosing what to do might be a good idea.

gabrielebndn:
> I am fine with #162. CMake computes a version number and configures a header. This is clean.
> 
> Talking about pinocchio, you now have the version number in the tags and in the package.xml so it is not better.
> 
> The question is about how our CMake module computes the version number. We use `git describe` for this but there are cases where it fails (and we should not assume this will always work). So we have several options:
> 
> * add a failback (like package.xml) but this means the version number will be into 2 places.
> * always read it from package.xml no matter what the tag is.
> * hard code it in the CMakeLists.
> * any options I may not have in mind.
> 
> To me, the best is to provide a macro for each solution and let the project decide what fits their needs.
> 
> What I am asking for is a clear discussion on that and not a sprint where developers are involved only if they answer within the next 2 hours a PR is opened.

No matter which failback we might add in order to provide a meaningful version, it will amount to hard-coding it somewhere, be it in `package.xml` or in the `CmakeLists.txt` If we are confident that we can manually keep track of the versions, then this whole automatic computation is useless.
And the truth is that we have serious difficulty keeping track of the versions, especially as releases are very frequent. By the way, we have already seen problems with `package.xml`, see https://github.com/stack-of-tasks/pinocchio/issues/916#issuecomment-546368356

nim65s:
I would be in favor of computing them from the git tag and the package.xml, and letting `make release` update both. This way, we could remove the `.version` file, as it would be redundant with the package.xml.

jmirabel:
> I would be in favor of computing them from the git tag and the package.xml, and letting `make release` update both. This way, we could remove the `.version` file, as it would be redundant with the package.xml.

Does this mean that every package should have a package.xml ?

gergondet:
I'm rather on @jmirabel side here. I think it's important to leave the choice to the package developer/maintainers. I don't want to add a `package.xml` in packages where that doesn't make sense.

Imho, the simplest/safest solution is to keep a single source of truth and if the information absolutely needs to be duplicated then it should be enforced by a CI check (e.g. if I opt for having `PROJECT_VERSION` defined but I need a `package.xml` then I can have my CI check that the two versions always matches.

A final note, one doesn't have to merge everything into master to use a new feature. You can push a branch on this repository and just update the reference on the consumer side. This will avoid rushed PRs to go through before they are properly tested. (you can also change the remote on the consumer side to point to your fork, that's annoying for regular users but in a CI/devel environment this doesn't matter quite as much)

nim65s:
I think the suggestion to put a `package.xml` in all packages originated from @wxmerkt ; do you want to add something here ? Or maybe I misunderstood something.

nim65s:
@gergondet : your suggestion about pushing branches is really good and would really help me, but I currently don't have the rights for this, can I get those ?

gergondet:
@nim65s done

gabrielebndn:
> I don't want to add a package.xml in packages where that doesn't make sense.

It could be made optional

> it should be enforced by a CI check

Having CI checks on this is a good idea.

jcarpent:
Anyway, we should not rely on git to describe the current version of the project. But maybe, having a standard version.cmake in the project core should be helpful for easy update by the `make release` command

olivier-stasse:
I proposed something similar to @wxmerkt for the package.xml (i.e. check the version during make release). We could have a ros branch adding a package.xml for those who do not want to have a package.xml in the main branch. 

The point for me is that package.xml is to be used in conjunction with a super build system for the user. Providing the ros branch with vcs import is easy, pushing back the modification of the user in the main branch seems to add a level of unnecessary complexity.
 
What do you think ?

jmirabel:
In which repo would you add a ros branch ?

wxmerkt:
> We could have a ros branch adding a package.xml for those who do not want to have a package.xml in the main branch.

This sounds like a cumbersome idea for maintenance to prevent a single file.

olivier-stasse:
Ok then we all agree that this file should be not put in a different branch.

PR #126, opened 2018-03-05
Change license of generated files from GPL / LGPL to BSD

jmirabel:
This module generates a few header files with the GPL license. They are installed and therefore force dependencies to be GPL.

This pull-request changes the license to BSD-2.

@thomas-moulard : as the author of these files, if you agree, could you merge this PR ?

PR #271, opened 2019-11-07
Remove license header

jcarpent:


jmirabel:
Any reason for this ?

jcarpent:
There is no real reason to propagate this license towards the packages that use jrl-cmake-modules

jmirabel:
First of all, the license is not *propagated*. It is restricted to the 3 files that are generated and does not influence in any way the license of the software.

The license here makes it explicit to people that they can do whatever they want with this file.

jcarpent:
I do think there is a reason to put a license on a generated files, but I would rather suggest to directly refer to the one from the sofware which uses jrl-cmake. 
I think this adding noise where there is no need.

jmirabel:
The message is not very explicit to that extend. You should likely add a statement like
```
///////////////////////////////////////////////////////////////////////////////
// BSD 2-Clause License
// Copyright (C)  2008-2018 LAAS-CNRS, JRL AIST-CNRS, INRIA. 
// See <some link>
///////////////////////////////////////////////////////////////////////////////
```

By the way, this project is LGPL so removing the license of these headers makes them LGPL, which is the opposite to https://github.com/jrl-umi3218/jrl-cmakemodules/pull/126

jcarpent:
We then just need this file follows the rules of the targeted project, would be simpler and less borring.

jmirabel:
a. You are not the author of the file so you cannot decide by yourself to change its license.
b. Can the file be orphaned and its authorship and license be modified at configure time ? That sounds weird to me.

jcarpent:
a. Indeed, you are correct with this respect.
BUT, I do think the intention of this file is to propagate a license which has nothing to do with the end-user project. 
I've removed this line, because when I was working on the version, I was really surprised to see these headers which were not instructive, and was leading to some confusion.
I think it is better to make these files generic enough, and to avoid putting license on it. I was much more in favor of linking them to the generated files, which seems the correct way of knowing their origin for further modifications.

But again, if you want to revert this PR, feel free to do it. But, I do not see any converging point.

jcarpent:
Or add the header you suggested, I would not stop you.

PR #272, opened 2019-11-07
config: correct PINOCCHIO_VERSION_UNKNOWN_TAG

jcarpent:


jcarpent:
@gabrielebndn This solution is working well, without changing the existing code!

header.cmake
```diff
@@ -147,20 +147,20 @@ ENDMACRO(_SETUP_PROJECT_HEADER)
 FUNCTION(GENERATE_CONFIGURATION_HEADER
     HEADER_DIR FILENAME LIBRARY_NAME EXPORT_SYMBOL)
 
-  IF(${PROJECT_VERSION_MAJOR} MATCHES UNKNOWN) 
-    SET(PROJECT_VERSION_MAJOR_CONFIG "${PROJECT_NAME}_VERSION_UNKNOWN_TAG")
+  IF(NOT ${PROJECT_VERSION_MAJOR} MATCHES UNKNOWN)
```
gabrielebndn: I'm sorry, why the `NOT`? I think it's a leftover, it should be removed, isn't it?

gabrielebndn:
Better. This implies that version numbers may be signed, which may cause issues if somebody assumes `unsigned int`

jcarpent:
Right, I would change then for 0, as you suggested, even if I'm favor in 0.0.0 version for first version.

PR #276, opened 2019-11-07
[config.hh] Add @LIBRARY_NAME@_VERSION_AT_LEAST

gabrielebndn:
This would save lots of boilerplate, and it would provide all of our packages with a uniform interface.


gabrielebndn:
@nim65s, what do you think?

nim65s:
@jmirabel this could be useful to handle deprecation in an even smoother way, so why not

gabrielebndn:
I do not see much need of it, but why not, I pushed `@LIBRARY_NAME@_VERSION_AT_MOST` too. If you do not spot any error, I think you can merge this PR

jmirabel:
Looks good to me.

PR #279, opened 2019-11-08
Pkgonfig spaces

sbarthelemy:
This pull request reverts
https://github.com/jrl-umi3218/jrl-cmakemodules/pull/265
and fixes the original issue (compatibility with perl implementation of pkg-config) differently.

It also adds some basic CI.


pkg-config.cmake
```diff
@@ -220,17 +220,30 @@ MACRO(_SETUP_PROJECT_PKG_CONFIG_FINALIZE)
...
+# - _PKG_CONFIG_STRING_NOSPACE_VAR <- `my-package>0.4`
+# `my-package` results in
+# - _PKG_LIB_NAME_VAR <- my-package
+# - _PKG_PREFIX_VAR <- MY_PACKAGE
+# - _PKG_CONFIG_STRING_NOSPACE_VAR <- `my-package`
+MACRO(_PARSE_PKG_CONFIG_STRING PKG_CONFIG_STRING _PKG_LIB_NAME_VAR _PKG_PREFIX_VAR _PKG_CONFIG_NOSPACE_VAR)
+  # Decompose the equation
+  STRING(REGEX MATCH "([^ ]+)( (>|>=|=|<=|<) (.*))?" _FOO "${PKG_CONFIG_STRING}")
+  # Reconstruct the equation, without the space around the operator
+  SET(${_PKG_CONFIG_NOSPACE_VAR} "${CMAKE_MATCH_1}${CMAKE_MATCH_3}${CMAKE_MATCH_4}")
```
jmirabel: I think the access to regex groups was added in a recent version of CMake. CMake 3.2 does not support it (at least it is not in the [doc](https://cmake.org/cmake/help/v3.2/command/string.html)).
You can work around it with `STRING(REGEX REPLACE ...)` which support regex groups.

_unittests/test_pkg-config.cmake
```diff
@@ -0,0 +1,34 @@
...
+EXPECT_STREQUAL("my-package" "${PKG_LIB_NAME}")
+EXPECT_STREQUAL("MY_PACKAGE" "${PKG_PREFIX}")
+EXPECT_STREQUAL("my-package>=0.4" "${PKG_CONFIG_STRING_NOSPACE}")
+
+UNSET_TEST_VARS()
+_PARSE_PKG_CONFIG_STRING("my-package" PKG_LIB_NAME PKG_PREFIX PKG_CONFIG_STRING_NOSPACE)
+EXPECT_STREQUAL("my-package" "${PKG_LIB_NAME}")
+EXPECT_STREQUAL("MY_PACKAGE" "${PKG_PREFIX}")
+EXPECT_STREQUAL("my-package" "${PKG_CONFIG_STRING_NOSPACE}")
+
```
jmirabel: Can you test the version without spaces too ?

pkg-config.cmake
```diff
@@ -220,17 +220,30 @@ MACRO(_SETUP_PROJECT_PKG_CONFIG_FINALIZE)
...
-  STRING(REGEX MATCH "[^<>= ]+" ${_PKG_LIB_NAME_VAR} "${PKG_CONFIG_STRING}")
-  # And transform it into a valid variable prefix.
+# - _PKG_CONFIG_STRING_NOSPACE_VAR <- `my-package>0.4`
+# `my-package` results in
+# - _PKG_LIB_NAME_VAR <- my-package
+# - _PKG_PREFIX_VAR <- MY_PACKAGE
+# - _PKG_CONFIG_STRING_NOSPACE_VAR <- `my-package`
+MACRO(_PARSE_PKG_CONFIG_STRING PKG_CONFIG_STRING _PKG_LIB_NAME_VAR _PKG_PREFIX_VAR _PKG_CONFIG_NOSPACE_VAR)
+  # Decompose the equation
+  STRING(REGEX MATCH "([^ ]+)( (>|>=|=|<=|<) (.*))?" _FOO "${PKG_CONFIG_STRING}")
```
jmirabel: Shouldn't it be `([^ ]+)( ?(>|>=|=|<=|<) ?(.*))?` ? So that it supports the case with no spaces.

Moreover, could you rename `_FOO` to `_UNUSED` ?

_unittests/test_pkg-config.cmake
jmirabel: Thanks.

pkg-config.cmake
sbarthelemy: The 3.2 version doc says

```
 ()        Saves a matched subexpression, which can be referenced
           in the REGEX REPLACE operation. Additionally it is saved
           by all regular expression-related commands, including
           e.g. if( MATCHES ), in the variables CMAKE_MATCH_(0..9).
```

I'll try to find a way to get can old cmake version on the CI.

_unittests/test_pkg-config.cmake
sbarthelemy: I planed to *not* support the version without space.
It was never supported by jrl-cmakemodules AFAIK, and it seems better to stick to pkg-config's behavior than to support FindPkgConfig.cmake's one.

pkg-config.cmake
sbarthelemy: > Shouldn't it be `([^ ]+)( ?(>|>=|=|<=|<) ?(.*))?` ? So that it supports the case with no spaces.

cfr above, my plan is not to add support for the case without spaces.

Is it five with you?

> Moreover, could you rename `_FOO` to `_UNUSED` ?
done.

Thanks a lot for the review!

pkg-config.cmake
jmirabel: You are right. I missed it. Sorry for the noise.

jcarpent:
By the way, the project does not support GitHub action yet.

pkg-config.cmake
jmirabel: > cfr above, my plan is not to add support for the case without spaces.

It might be both surprising and annoying for projects that does `ADD_REQUIRED_DEPENDENCY("foo>1"))`. Do you know what error message they will get ?

If it is something like "*package 'foo>1' was not found*" then it is really tricky because it is fairly hard to understand that the issue is the missing space...

pkg-config.cmake
sbarthelemy: > I'll try to find a way to get an old cmake version on the CI.

- ubuntu 16.04 ships with cmake 3.5.
- [github actions' ubuntu 16.04](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners#ubuntu-1604-lts) ships with cmake 3.12.
- [The pip package](https://pypi.org/project/cmake/) seems to start at [3.6](https://pypi.org/project/cmake/0.1.0/)


If old cmake versions need to be tested, the simplest solution might be yo use the kitware binaries (eg. https://cmake.org/files/v3.2/)

Anyway, there is no hurry as long as the project has no CI in place.

pkg-config.cmake
sbarthelemy: > > cfr above, my plan is not to add support for the case without spaces.
> 
> It might be both surprising and annoying for projects that does `ADD_REQUIRED_DEPENDENCY("foo>1"))`. Do you know what error message they will get ?

without this pull request, the project would build with success, but generate an invalid ".pc".
dependers would fail at cmake run with

```
CMake Error at /usr/share/cmake-3.10/Modules/FindPkgConfig.cmake:415 (message):
  A required package was not found
  /usr/share/cmake-3.10/Modules/FindPkgConfig.cmake:593 (_pkg_check_modules_internal)
  cmake/pkg-config.cmake:298 (PKG_CHECK_MODULES)
  cmake/pkg-config.cmake:510 (ADD_DEPENDENCY)
  CMakeLists.txt:36 (ADD_REQUIRED_DEPENDENCY)
```
(this just happened to pinocchio dependers who use devel, see
https://github.com/stack-of-tasks/pinocchio/pull/933#issuecomment-551333575)

I can try to build example-robot-data next week and see if this commit would change anything with this respect.

> If it is something like "_package 'foo>1' was not found_" then it is really tricky because it is fairly hard to understand that the issue is the missing space...

I agree, I don't think this pull request makes the matter worse, but I can check next week.

By the way, PKG_CHECK_MODULES' implementation seems incomplete. If I read it correctly, it would parse "toto>=3" and convert it into "toto >=3", but not "toto>3". (I don't think we need to do anything about it)

jmirabel: > without this pull request, the project would build with success, but generate an invalid ".pc".

Thanks for the details. I don't think you have to worry more than this.

PR #282, opened 2019-11-12
Update computation of version method.

jmirabel:
This follows the discussions in #270 

Here is a draft to continue the discussion. I will update the documentation once we agree on the method.

This adds 3 functions `_COMPUTE_VERSION_FROM_DOT_VERSION_FILE`, `_COMPUTE_VERSION_FROM_GIT_DESCRIBE` and `_COMPUTE_VERSION_FROM_ROS_PACKAGE_XML_FILE`. I hesitated on the naming. I put `_` in the beginning to make them somehow *hidden*. However, if we want them to be used directly by the user, then they should be prefixed with `PROJECT_...`

The macro `VERSION_COMPUTE` reads a variable called `PROJECT_VERSION_COMPUTATION_METHODS` and calls the desired computation methods in the desired order. By default, it uses the .version, then git describe.

If a package needs a different option, it can for instance do `SET(PROJECT_VERSION_COMPUTATION_METHODS "ROS_PACKAGE_XML_FILE")`. If only one method is desired, then it may be clearer to call `_COMPUTE_VERSION_FROM_ROS_PACKAGE_XML_FILE` (hence the hesitation about the function naming).

Any opinion on this approach ?


gergondet:
I think that's a good idea that leaves version management up to the maintainers of the package.

For documentation purpose I think it's important to indicate that setting `PROJECT_VERSION` (prior to `include(cmake/base.cmake)`) will override the mechanism entirely.

> If only one method is desired, then it may be clearer to call `_COMPUTE_VERSION_FROM_ROS_PACKAGE_XML_FILE` (hence the hesitation about the function naming).

I am not sure I follow what you mean here. Do you mean the user should call the function by himself? Right now, `COMPUTE_VERSION()` is automatically called by including `cmake/base.cmake` so I'm not sure how that would work

nim65s:
I think we could do that automatically:

1. if a `package.xml` file exists, just get the version from here
2. else, if a `.version` exists, get the version from there
3. else, if a tag exists, get the version there
4. else, version is undefined.

Version tags should be created only with a `make release`, and the `make release` should either update the version in `package.xml` if it exists, or create a `.version` otherwise.

With this, I think we can cover all our use cases.

jmirabel:
> I am not sure I follow what you mean here. Do you mean the user should call the function by himself? Right now, `COMPUTE_VERSION()` is automatically called by including `cmake/base.cmake` so I'm not sure how that would work

Yes, it would me to include version.cmake beforehand. Then, `COMPUTE_VERSION()` is automatically called only if PROJECT_VERSION is not defined.

jmirabel:
@nim65s you mean that the default value of `PROJECT_VERSION_COMPUTATION_METHODS` should be `"ROS_PACKAGE_XML_FILE" "DOT_VERSION_FILE" "GIT_DESCRIBE"` ?

I think the default value should be as simple as possible (with the constraints of backward compatibility ?). Then, complex behaviors in individual package can be implemented by setting the above variable.

nim65s:
This would be a nice way of implementing that order, yes, and I think this would allow backward compatibility for projects that don't have a `package.xml` file, as the `.version` files already had precedence over the git tags (as this already hit me: https://github.com/stack-of-tasks/dynamic-graph/commit/9a99f20af79edeec02e7360f05219e62bc6bf06f).

Then, for project that don't want a `package.xml` files, there will be no changes wrt to the old behavior, and for those who do, it wakes sense to use it first by default.

I think this would be the simplest way to get what we want everywhere, if we look at how complex is it to implement this here, and update all the packages that use this module with a new cmake configuration.

nim65s:
Plus… As i'm maintaining a lot of projects, I really prefer when they are all behaving in the same manner, without having to update all of them x)

olivier-stasse:
Then it would be mean that for make release you do not give the tag in argument but your read from the file (package.xml, or .version) ?

Or for make release only the argument matters, and then you make sure:
 *  that package.xml is consistent if it exists, 
 *  .version is consistent if it exists.

Having the .version in the main repository already tricked me in the past.

jcarpent:
I share the same opinion as @nim65s to centralize the localization of the version. I do not think that the `.version` is the right place as it is non-standard. I do believe that a generic `package.xml`(inspired by ROS) is a good option.
Finally, I do think that `make release` should:
- commit the update of the file containing the current version number
- tag the git tree with the version (for easy release on GitHub)

nim65s:
@olivier-stasse : I think we should keep the `make release VERSION=x.y.z`, and this make target should handle the update of the version in `package.xml` (which implies a git commit) if it exists. If it doesn't, then we can continue to use the usual `.version`, but only inside the release tarball.

@jcarpent : I agree, but [concerns were raised](https://github.com/jrl-umi3218/jrl-cmakemodules/pull/270#issuecomment-551644246) about requiring a `package.xml` file, this is why I proposed a solution where this is the default thing to use and update if it exists, but keep the old behavior if it doesn't.

gabrielebndn:
I agree on having one or more files which are automatically updated by `make release` at the same time the version tag is put and and which also takes care of committing the change, if and only if they exist, so that nobody is forced to put them if they do not want them. The automatic commit should have a descriptive commit message, such as "Version [...]" Then normal `make` should read those files, and only use `git describe` as last resort.

If more than one file is supported, then there should be a hierarchy, but they should *all* be updated if more than one exists in the specific project.

jmirabel:
From the discussion, I can't say whether a consensus was reached. Should this be merged or closed ?

nim65s:
For me, this can be merged.

I would prefer `PROJECT_VERSION_COMPUTATION_METHODS` to default to `"ROS_PACKAGE_XML_FILE" "DOT_VERSION_FILE" "GIT_DESCRIBE"`, because if a package has a `package.xml` file, it would be counter-intuitive and error-prone not to use it by default, but to get this work properly, we would have to add an update of this file and a git commit in the `make release` target, which can be added later.

jmirabel:
> I would prefer `PROJECT_VERSION_COMPUTATION_METHODS` to default to `"ROS_PACKAGE_XML_FILE" "DOT_VERSION_FILE" "GIT_DESCRIBE"`

Could be. I'll wait for other comments.

> because if a package has a `package.xml` file, it would be counter-intuitive and error-prone not to use it by default, but to get this work properly, we would have to add an update of this file and a git commit in the `make release` target, which can be added later.

I would prefer avoiding that discussion here and let this discussion happen project-wise.

wxmerkt:
Ping - what is the status of this?

olivier-stasse:
ping again

jcarpent:
On my side, I'm OK for merging. We should also provide an automatic update of package.xml with the new version.

jmirabel:
There is still one pending question: what should be the default value of `PROJECT_VERSION_COMPUTATION_METHODS`. @nim65s suggested

> I would prefer `PROJECT_VERSION_COMPUTATION_METHODS` to default to `"ROS_PACKAGE_XML_FILE" "DOT_VERSION_FILE" "GIT_DESCRIBE"`

jcarpent:
Yes, the suggestion of @nim65s is fine for me.

jcarpent:
@jmirabel Would you have time to finish this PR. I think with it we will make ROS happy ;)

jmirabel:
Done.

jmirabel:
I will merge this. Please, mind that, next time you update the CMake module, the version will by default first be looked for in the package.xml. This means the *git describe* suffix won't be there.

PR #285, opened 2019-11-18
Changes related to exporting targets.

### LINKED ISSUE #283: CMake warning about CMP0007
This line
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/8e01cbbb07722838d69d098ae6abb4211fae1737/cmake_uninstall.cmake.in#L17

added by #52 generates the following warning
> CMake Deprecation Warning at cmake/cmake_uninstall.cmake:17 (cmake_policy):
> The OLD behavior for policy CMP0007 will be removed from a future version of CMake.
>
> The cmake-policies(7) manual explains that the OLD behaviors of all policies are deprecated and that a policy should be set to OLD only under specific short-term circumstances.  Projects should be ported to the NEW behavior and not rely on setting a policy to OLD.

I think adding `LIST(REMOVE_ITEM files "")` after 
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/8e01cbbb07722838d69d098ae6abb4211fae1737/cmake_uninstall.cmake.in#L35
should be sufficient.

jmirabel:
This PR also:
- In the pkg-config dependencies, use the cmake target macros so that the compilation options are also exported.

  This may not be backward compatible due to the `INCLUDE_DIRECTORIES` which becomes `TARGET_INCLUDE_DIRECTORIES`. It is not more available for all the target in a directory and its sub-directories. I don't think this will create much trouble.

- it exports the pkg-config dependencies to the config file. (I don't think this is necessary)
- it adds macro `COMPUTE_PROJECT_ARGS` to compute the arguments according to the version of cmake.
- it updates the documentation.
- it fixes #283


base.cmake
```diff
@@ -237,3 +243,29 @@ MACRO(SETUP_PROJECT_FINALIZE)
...
+    SET(${VARIABLE} VERSION ${PROJECT_VERSION_FULL} LANGUAGES CXX)
+
+    # Append description for CMake >= 3.9
+    IF(CMAKE_VERSION VERSION_GREATER "3.9.0")
+      SET(${VARIABLE} ${VARIABLE} DESCRIPTION ${PROJECT_DESCRIPTION})
+    ENDIF(CMAKE_VERSION VERSION_GREATER "3.9.0")
+  ELSE(CMAKE_VERSION VERSION_GREATER "3.0.0")
+    SET(${VARIABLE} CXX)
+  ENDIF(CMAKE_VERSION VERSION_GREATER "3.0.0")
+ENDMACRO(COMPUTE_PROJECT_ARGS)
```
jmirabel: This avoids having to write stuff like https://github.com/humanoid-path-planner/hpp-fcl/pull/105/commits/3b473bc13cc1bf2f2b70021e79f76f19cb01887f in every project.

gergondet: That's great but could we introduce `PROJECT_LANGUAGES` (defaults to `CXX`) in order to handle the case when the project does not only uses C++.

gergondet:
> This may not be backward compatible due to the INCLUDE_DIRECTORIES which becomes TARGET_INCLUDE_DIRECTORIES. It is not more available for all the target in a directory and its sub-directories. I don't think this will create much trouble.

It becomes available when you link with the target so it should be fine in most cases

I'm just wondering if it's ok to mix `target_include_directories` with the PUBLIC keyword and `target_link_libraries` without (I guess it is but did you try it in hpp-fcl for example?)

jmirabel:
I tried it on hpp-fcl and it worked. As far as I understand the doc of `target_link_library`, I think
`target_link_library(a b)` and `target_link_library(a PUBLIC b)` are the same. I will add the PUBLIC keyword anyway to keep consistency.

gergondet:
> I tried it on hpp-fcl and it worked. As far as I understand the doc of target_link_library, I think
target_link_library(a b) and target_link_library(a PUBLIC b) are the same.

They are but CMake gets very mad if you mix both. I guess it's fine to have keyword version

base.cmake
```diff
@@ -246,26 +246,35 @@ ENDMACRO(SETUP_PROJECT_FINALIZE)
...
+#       COMPUTE_PROJECT_ARGS(PROJECT_ARGS LANGUAGES CXX)
 #       PROJECT(${PROJECT_NAME} ${PROJECT_ARGS})
 #
-#     :param VARIABLE: the variable where to write the result
+#     :param OUTPUT_VARIABLE: the variable where to write the result
+#     :param LANGUAGES: the project languages. It defaults to CXX.
 #
-MACRO(COMPUTE_PROJECT_ARGS VARIABLE)
+MACRO(COMPUTE_PROJECT_ARGS)
+  CMAKE_PARSE_ARGUMENTS(_project_ "" "VARIABLE" "LANGUAGES" ${ARGN})
```
gergondet: It should be `_project` here and not `_project_`

see [doc](https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html#command:cmake_parse_arguments)

> When done, cmake_parse_arguments will consider for each of the keywords listed in <options>, <one_value_keywords> and <multi_value_keywords> a variable composed of the given <prefix> followed by "_" and the name of the respective keyword

jmirabel: Good catch. Thanks.

gergondet:
This looks ok to me. Feel free to merge :-)

(disregard CI, it won't be fixed until we can run GitHub Actions on this repo)

jmirabel:
I tested it on hpp-fcl and had to make a few changes to get it to work. It is now working properly.

PR #289, opened 2019-11-23
Add optional argument to TARGET_LINK_BOOST_PYTHON

jcarpent:


boost.cmake
```diff
@@ -193,18 +193,23 @@ MACRO(SEARCH_FOR_BOOST)
...
+#   :private,public,interface: The PUBLIC, PRIVATE and INTERFACE keywords can be used to specify both the link dependencies and the link interface.
 #
 #   On darwin systems, boost_python is not linked against any python library.
 #   This linkage is resolved at execution time via the python interpreter.
 #   We then need to stipulate that boost_python has unresolved symbols at compile time for a library target.
 #   Otherwise, for executables we need to link to a specific version of python.
 #
 MACRO(TARGET_LINK_BOOST_PYTHON target)
+  IF(${ARGC} GREATER 1)
+    SET(PUBLIC_KEYWORD ${ARGV1})
```
jmirabel: Won't this mess up with the current usage of PUBLIC_KEYWORD ?

jcarpent: I don't think this should be manager globally, but rather locally.

jmirabel: I agree with you.
But at the moment, there are still places where the global variable is accessed. We should modify them so that they don't use the global variable anymore.

PR #291, opened 2019-11-25
[Python] Look for python 3 by default

nim65s:
Hi,

Right now, we are incitating users to switch to Python 3. It would be way easier if they didn't had to type `-DPYTHON_EXECUTABLE=$(which python3)` each and every time.

But for those who still really want to stick with Python 2, `-DPYTHON_EXECUTABLE=$(which python2)`, or `FINDPYTHON(2)` should still be available.

This PR makes projects using `FINDPYTHON()` look for Python 3 first.

Also, this is now the standard behavior in modern CMake versions:
https://cmake.org/cmake/help/v3.16/module/FindPython.html


wxmerkt:
I realise this is an open discussion, however, I would not force/incite users to switch immediately despite January 1, 2020 approaching. All LTS distros will maintain their default Python until the end of lifetime of their distro. As such, the default Python should be the system default (or configured default).

python.cmake
```diff
@@ -55,7 +55,11 @@ ENDIF(CMAKE_VERSION VERSION_LESS "3.2")
 
 MACRO(FINDPYTHON)
 
-FIND_PACKAGE(PythonInterp ${ARGN})
+IF(ARGN)
+  FIND_PACKAGE(PythonInterp ${ARGN})
+ELSE(ARGN)
+  FIND_PACKAGE(PythonInterp 3)
```
wxmerkt: Please do not amend default behaviour on released LTS distros. LTS distros will maintain their default (and thus 2.7) till the end of distro life.

jmirabel: I did not think about that. I think @wxmerkt is correct.

nim65s:
As you prefer :)

PR #293, opened 2019-11-28
WIP: [xacro] use `PYTHON_EXECUTABLE` and fix `OUTPUT_NAME`

nim65s:
Hi,

I had to do this to get xacro working with python3, but I would understand that we don't need to use the same python interpreter to use xacro and for the rest of the project, so this is open for discussions.


xacro.cmake
```diff
@@ -87,5 +87,5 @@ MACRO(GENERATE_URDF_FILE FILENAME EXTENSION)
     SET(_XACRO_FILE_ ${CMAKE_CURRENT_SOURCE_DIR}/${FILENAME}.xacro)
   ENDIF (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/${FILENAME}.xacro)
 
-  RUN_XACRO(${_XACRO_FILE_} ${CMAKE_CURRENT_BINARY_DIR}/${OUTPUT_FILE})
+  RUN_XACRO(${_XACRO_FILE_} ${CMAKE_CURRENT_BINARY_DIR}/${FILENAME}.${EXTENSION})
```
jmirabel: How the hell did this ever worked !

xacro.cmake
```diff
@@ -40,8 +40,8 @@ MACRO(RUN_XACRO INPUT OUTPUT)
 
   ADD_CUSTOM_COMMAND(
     OUTPUT ${OUTPUT}
-    COMMAND ${_xacro_py}
-    ARGS ${XACRO_OPTIONS} -o ${OUTPUT} ${INPUT}
+    COMMAND ${PYTHON_EXECUTABLE}
```
jmirabel: Could you add a FINDPYTHON at the beginning of the macro ?

jmirabel:
I am not sure whether this is a good idea.
What issue did you get ?

xacro.cmake
nim65s: This is an excellent question. I guess a commit forgot to update this too at some point.

nim65s: or maybe xacro is able to understand that when the output is a directory, it just has to put a file with `${FILENAMEE}.urdf` in this directory.

nim65s:
I installed xacro from source from robotpkg while robotpkg was configured in python3, and xacro's main script started with `#!/usr/bin/env python`, and contained an `import xacro`, but the `xacro` python module was nowhere in the `sys.path` of the main `python` of my system, which was python2.

So, more generally, xacro can be installed with a python version which is independent from the OS, and raise this issue… But it can also (and this might occur more often) be installed with a python version different from `PYTHON_EXECUTABLE`, so maybe we would need a `XACRO_PYTHON_EXECUTABLE`, but this looks overkill.

I don't have any good solution right now, this is why I just wanted to get opinions.

xacro.cmake
nim65s: Sure.
Will do if we decide do keep it :)

nim65s:
rebased

nim65s:
#348 was included by mistake during the rebase ; please don't merge :)

nim65s:
Actually, this works well: https://github.com/ros/xacro/blob/1.13.5/cmake/xacro-extras.cmake

eg.
```cmake
FOREACH(F ${XACRO_FILES})
  xacro_add_xacro_file(${F}.urdf.xacro)
  INSTALL(FILES ${XACRO_OUTPUT_FILE} DESTINATION …)
ENDFOREACH(F ${XACRO_FILES}) 
```

So I suggest to deprecate this xacro.cmake file :)

jmirabel:
> ```cmake
> ```cmake
> 
> ```
> 
> 
> So I suggest to deprecate this xacro.cmake file :)
> ```

Agreed.

nim65s:
replaced by #490

PR #299, opened 2019-12-04
[Sphinx] Install doc only if INSTALL_DOCUMENTATION is TRUE

jmirabel:


nim65s:
This has been in my todolist for more than a year… Thanks for handling it !

jcarpent:
Just to know, the sphinx module does it allow to install documentation of Python bindings?

jmirabel:
> Just to know, the sphinx module does it allow to install documentation of Python bindings?

Yes, it does.

jcarpent:
Thanks a lot for the quick reply. Could you provide an example of use? I will include it to Pinocchio ;)

jmirabel:
Have a look at sot-core/doc/CMakeLists.txt.

PR #301, opened 2019-12-05
Create reinstall target

### LINKED ISSUE #298: Make reinstall
When doing `make install`, the old version is kept there (particularly for the SOT): the new version is simply added next to the old version and then the symlinks are updated (see https://github.com/jrl-umi3218/jrl-cmakemodules/issues/189).

As an alternative to the propositions in https://github.com/jrl-umi3218/jrl-cmakemodules/issues/189 (which were ultimately rejected) I propose the following: an additional target `make reinstall` which will just run `make uninstall` first and then `make install`. This simple shortcut should ease the development.

Would you agree on that?

gergondet:
Closes #298

This is implemented so that `make reinstall` is also working on the first time it's invoked or if the project has already been uninstalled.

For this to work correctly, `make uninstall` also removes the install manifest

### DIFF
--- a/cmake_reinstall.cmake.in
+++ b/cmake_reinstall.cmake.in
@@ -1,0 +1,19 @@
+# Copyright (C) 2019 LAAS-CNRS, JRL AIST-CNRS, INRIA.
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


gabrielebndn:
Nice, thanks a lot!

Issue #306, opened 2019-12-17
Cyclic symlink in cmake->_unittests->cmake


blackliner:
Hi, i am trying to use eigen-qld within a bazel-built project (by using [rules_foreign_cc](https://github.com/bazelbuild/rules_foreign_cc)'s cmake_external) and bazel crashes because of that symlink cycle.

Is that symlink absolutely necessary? Cheers

gergondet:
Thanks for reporting this.

It's only used as part of the project's CI so it should be fine to remove it if it's created by the corresponding action. Would you have time to propose a PR for that?

blackliner:
I'll give my best :D

blackliner:
Done

blackliner:
Done, added the following to the CI script>
`ln -s .. cmake`

gergondet:
Closed by #307

PR #307, opened 2019-12-17
Not needed symlink removed

blackliner:
As discussed in https://github.com/jrl-umi3218/jrl-cmakemodules/issues/306


gergondet:
@blackliner you need to update `.github/workflows/cmake.yml` so that it creates the symlink :)

gergondet:
Thanks @blackliner

PR #308, opened 2019-12-17
CMake package: update example

nim65s:
As doc above explicitely says we must call this
`SETUP_PROJECT_PACKAGE_FINALIZE()`


gergondet:
Actually, you must also have `set(PROJECT_USE_CMAKE_EXPORT TRUE)` otherwise you need to call `setup_project_package_finalize()`

PR #309, opened 2019-12-17
fix uninstall for CMake 2.8, fix #305

### LINKED ISSUE #305: `$<CONFIGURATION>` use in uninstall.cmake
Hi,

In #301, there is a  `$<CONFIGURATION>`:
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/349e9bcac4a3006b61d5eff2956d3cf2f3595dce/uninstall.cmake#L42

According to [CMake docs](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#id2), this is deprecated since CMake 3.0, and we should use `$<CONFIG>`.

And I don't understand what this is meant to do. In the docs, I can only find examples with eg. `$<CONFIG:Debug>`, but in my CMake 3.10.2 (18.04), it works only with `CONFIG` and not `CONFIGURATION`.

Also, in CMake 2.8.12 (14.04), the current version fails at configure time with:
```
CMake Error in CMakeLists.txt:
Evaluation file
"…/eigenpy-1.6.12/cmake/Release/cmake_reinstall.cmake"
cannot be written.
```

nim65s:


gergondet:
Thanks for the fix @nim65s

PR #311, opened 2019-12-19
Python: remove bugged libs hints, fix #310

### LINKED ISSUE #310: Issue in PYTHON_LIBRARY_HINT
Hi,

https://github.com/jrl-umi3218/jrl-cmakemodules/blob/54ece258eed16da94c7f10979588fcb47b744eb8/python.cmake#L68-L75

I don't understand what is the purpose of this `PYTHON_LIBRARY_HINT` variable, but it seems that it is not used anywhere, and we even have a `TODO: Should be removed later` here:

https://github.com/jrl-umi3218/jrl-cmakemodules/blob/54ece258eed16da94c7f10979588fcb47b744eb8/python.cmake#L98

Moreover, it looks like in cases where the python command fails, the `ERROR_QUIET` implies that the `PYTHON_LIBRARY_HINT` doesn't exist, and therefore the `string(REGEX REPLACE …)` is wrong because it doesn't have the right number of defined arguments:

```
CMake Error at cmake/python.cmake:75 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.
  CMakeLists.txt:29 (FINDPYTHON)


CMake Error at /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find PythonLibs (missing: PYTHON_INCLUDE_DIRS) (found version
  "3.6.9")
  /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.10/Modules/FindPythonLibs.cmake:262 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  cmake/python.cmake:97 (FIND_PACKAGE)
  CMakeLists.txt:29 (FINDPYTHON)
```

This error appears randomly, and we (@florent-lamiraux & I) don't know how to reliably reproduce the error in this `EXECUTE_PROCESS` that seems to cause the issue, because once again, this `ERROR_QUIET` doesn't help to debug.

Suggestion: drop that code.
Ref #262 & #263, where this code was updated, but without clear description or context.

nim65s:


jmirabel:
Given the discussion in #310, I merge this.

PR #314, opened 2020-01-21
Making doxygen configuration fully available from CMakeLists

aescande:
This PR reworks the files `doxygen.cmake` and `Doxyfile.in` to allow for changing any doxygen flag from a `CMakeLists` file.

While doxygen can be fine-tuned from `doc/Doxyfile.extra.in`, this allows to have a coherent behavior within the `CMakeLists` files, where a handful of (kind of random) flags can be overridden, such as `DOXYGEN_INPUT`, `DOXYGEN_DOT_IMAGE_FORMAT` or `DOXYGEN_USE_MATHJAX`, while setting others will silently do nothing.

The PR 
 - adds a CMake macro, `_SETUP_DOXYGEN_DEFAULT_OPTIONS`, based on the `CMakeDoxygenDefaults.cmake` file that CMake can generates, that sets default values for all options not user-specified,
 - ensures that every option setting in the Doxyfile has the form OPTION = @DOXYGEN_OPTION@.
Default values are taken from the current Doxyfile.in

The PR consists in 4 commits:
 - 31cda80 makes most of the available options configurable (with the exception of options that were already derived from CMake variables). The Doxyfile generated after this commit is quasi-identical to the file that used to be generated. The differences are ignored by doxygen. Those are: one additional space after some `=`, additional quotes for some fields, minor reformatting or update of the comments and a slightly different syntax for the fields `ABBREVIATE_BRIEF` and `FILE_PATTERNS` where I do as in the `CMakeDoxygenDefaults.cmake`
 - f31ecd1 adds additional options that could be found in `CMakeDoxygenDefaults.cmake`
 - 2639c4e deals with the options that were already derived from CMake variables
 - bc521a7 adds all options that appears in the latest doxygen documentation


jmirabel:
This is a useful feature. Thanks for this.

I have one remark on its behavior. I would prefer something like:
1. Use `doxygen -g Doxyfile.default` to generate a config file with the default values,
2. From cmake, create a `Doxyfile.cmake` that contains the lines `${VARIABLE} = ${DOXYGEN_${VARIABLE}}` for each *VARIABLE* for which *DOXYGEN_${VARIABLE}* is defined.
3. Eventually, concatenate * Doxyfile.default*, *Doxyfile.cmake* and *Doxyfile.extra*.

What annoys me is that the default values are stored within this CMake code. I would prefer them to be given by doxygen itself.

aescande:
I see your point, and that would be a clean way to go. The path I chose was in the direct continuation of our use of a `Doxyfile.in`.
From personal preferences, one disadvantage of your approach is that all non-default options are at the end of the `Doxyfile` and away from the comments documenting their behaviors, meaning that when looking at the file to understand or tweak doxygen behavior, you would need quite a lot of back and forth within the file or between the file and the doxygen documentation. The current solution has the advantage of having a single statement per option (but for the options in doxyfile.extra) and the explanation of the option immediately above. 
Right now, we have 35-40 options that differs from the default Doxyfile (assuming `CMakeDoxygenDefaults.cmake` indeed contains the default options).

Now if I change to your approach, I have a few remarks and questions:
 - we'll still need a sizable `_SETUP_DOXYGEN_DEFAULT_OPTIONS` for the above mentioned 35-40 options
 - for some fields, quotes are currently added around the value within the `Doxyfile.in`. Some were added along the years (e.g. for `PERL_PATH` or `DOT_PATH`), other I added in this PR to prevent possible problems. This is done on a per-case basis, and I would need to translate that where I generate the lines `${VARIABLE} = ${DOXYGEN_${VARIABLE}}`. Putting quotes on every field value does not seem to be an option as some fields accept lists of values
 - is there a place in our cmake code were the `Doxyfile` derived from `Doxyfile.in` and the `Doxygen.extra` are concatenated? I can't find it. Where would you see the concatenation you propose occur?

jmirabel:
>     * for some fields, quotes are currently added around the value within the `Doxyfile.in`. Some were added along the years (e.g. for `PERL_PATH` or `DOT_PATH`), other I added in this PR to prevent possible problems. This is done on a per-case basis, and I would need to translate that where I generate the lines `${VARIABLE} = ${DOXYGEN_${VARIABLE}}`. Putting quotes on every field value does not seem to be an option as some fields accept lists of values

I had not thought about that.

> 
>     * is there a place in our cmake code were the `Doxyfile` derived from `Doxyfile.in` and the `Doxygen.extra` are concatenated? I can't find it. Where would you see the concatenation you propose occur?

At the end of `doxygen.cmake`.

I am fine with the PR as it is.

jmirabel:
@aescande I added some documentation at the beginning of the file. Feel free to modify it. If you want to see the result, refers to https://jrl-cmakemodules.readthedocs.io/en/master/pages/developers.html#update-the-doc

aescande:
Thanks for adding that. That's good for me.

PR #316, opened 2020-01-22
Use new find_package for CMake >= 3.12 to search for Python

jcarpent:


python.cmake
```diff
@@ -349,7 +375,7 @@ MACRO(PYTHON_INSTALL_BUILD MODULE FILE DEST)
 
   MESSAGE(AUTHOR_WARNING "PYTHON_INSTALL_BUILD is deprecated and will be removed in the future")
   MESSAGE(AUTHOR_WARNING "Please use PYTHON_INSTALL_ON_SITE")
-  MESSAGE(AUTHOR_WARNING "ref https://github.com/jrl-umi3218/jrl-cmakemodules/issues/136")
+  MESSAGE(AUTHOR_WARNING "ref https://github.com/jrl-umi3.12/jrl-cmakemodules/issues/136")
```
wxmerkt: I don't think this change is intended

jcarpent: Yes, you're right! Thanks.

jcarpent: Yes, you're right! Thanks.

PR #317, opened 2020-01-22
python: fix issue when looking for Python2

jcarpent:
Indeed FIND_PACKAGE(Python ...) looks first for Python 3.


wxmerkt:
So this defaults to Python2 unless Python3 requested?

jcarpent:
No. If you don't provide any clue, it will look for Python3 ;)

jcarpent:
In fact, there is no way to force `FIND_PACKAGE(Python ...)` to look first for Python2. Then, you have to force it by calling: `FIND_PACKAGE(Python2 ...)`.

wxmerkt:
So if the previous PR changed the default behaviour, that would be bad. Is the default behaviour from before preserved?

jcarpent:
Yes, it changed. So, but the previous behavior was not deterministic neither. There is no mention in CMake < 3.12 that cmake is looking first for Python 2 then for Python 3.

CMake >= 3.12 looks by default for Python3. Knowing that Python2 is now deprecated, we should enforce this policy, no?
@nim65s @jmirabel @gergondet What do you think?

jcarpent:
I checked and everything is going fine. The FIND_PACKAGE(Python ...) is then looking in the PYTHONPATH. If this one is pointing to Python2, then it tries to find it.

wxmerkt:
> Knowing that Python2 is now deprecated, we should enforce this policy, no?

It's not deprecated on LTS platforms and will be supported until the LTS support runs out. We should not change defaults.

PR #321, opened 2020-01-23
update GNUInstallDirs.cmake

nim65s:
This is especially useful to stop having to override `CMAKE_INSTALL_LIBDIR`, because we get a `lib64` default everywhere, even if nobody has a 32 bits system anymore.

Another option would be to stop shipping this file, which is provided by CMake >= 2.8.5, but the commit message from 884e3bd4 lacks explicity about the reasons of the inclusion of this file.

### DIFF
--- a/GNUInstallDirs.cmake
+++ b/GNUInstallDirs.cmake
@@ -1,72 +1,180 @@
-# - Define GNU standard installation directories
-# Provides install directory variables as defined for GNU software:
-#  http://www.gnu.org/prep/standards/html_node/Directory-Variables.html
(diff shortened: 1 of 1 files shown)


olivier-stasse:
The explanation was simply to apply GNU rules in term of directory naming instead of having a per user rule. IMHO if there is an official inclusion in cmake then this file could be removed only if this does not break in other systems. AFAIK it seems that not all OS are consistent in terms of cmake receipes. Thus it will be safer to apply your patch until we check that removing the file does not cause any problem.

Thanks for the update.

PR #325, opened 2020-02-19
python: fix comptibility with previous search policy

jcarpent:
@wxmerkt I think this PR fixes the issue you mentioned in Pinocchio and EigenPy.


wxmerkt:
Thank you for the quick turnaround :-) This works well with Crocoddyl

nim65s:
Actually, this fix @wxmerkt current symptoms, but not the main problem.

I mean, what we want is CMake to find the python interpreter by default of the system.
My guess is that we should use `find_program(PYTHON_EXECUTABLE python)` or something similar. On systems with python 2 by default, this should give python 2, and on systems with python 3 by default, this should give python 3

PR #326, opened 2020-02-20
Search for Python executable if no hints provided

jcarpent:
Following discussions in #325.


gergondet:
FYI, `FIND_PACKAGE("Python3" REQUIRED COMPONENTS Interpreter Development)` will return the latest Python version available so if you happen to have `/usr/bin/python3.7` and `/usr/bin/python3.8` but `/usr/bin/python` is `/usr/bin/python3.7` (this was the case in focal one/two months ago then `find_package(Python3 REQUIRED COMPONENTS Interpreter Development)` will give you `python3.8` (and related headers/libs). Fortunately, this issue won't show up anymore on focal since they have switched `python3` to `python3.8`

I have a work around here: https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/cython/cython.cmake#L297 to find the exact version corresponding to the interpreter. It relies on some undocumented variables of the find Python script

I hope this can save you some of the trouble I had to figure this out :D

jcarpent:
> I have a work around here: https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/cython/cython.cmake#L297 to find the exact version corresponding to the interpreter.

@gergondet I don't think I've totally understood the trick behind. :(

gergondet:
@jcarpent 

The meat is in:
```cmake
set(${PYTHON}_FIND_VERSION_COUNT 3)
set(${PYTHON}_FIND_VERSION_EXACT TRUE)
execute_process(COMMAND ${PYTHON_B} -c "import sys; print(sys.version_info.major);" OUTPUT_VARIABLE ${PYTHON}_FIND_VERSION_MAJOR OUTPUT_STRIP_TRAILING_WHITESPACE)
execute_process(COMMAND ${PYTHON_B} -c "import sys; print(sys.version_info.minor);" OUTPUT_VARIABLE ${PYTHON}_FIND_VERSION_MINOR OUTPUT_STRIP_TRAILING_WHITESPACE)
```

In this code, `${PYTHON_B}` is either `python`, `python2` or `python3` (as I have options to build Python2 and Python3 modules at the same time or to enforce one or the other over the default system `python`) and `${PYTHON}` is either `Python2` or `Python3`.

Line by line:

```cmake
set(${PYTHON}_FIND_VERSION_COUNT 3)
```

This line (if I remember correctly) is a dirty way to trigger some logic in the `find_package(Python2|Python3 ...)` code.

```cmake
set(${PYTHON}_FIND_VERSION_EXACT TRUE)
```

This is straight-forward I think, we request a specific version.

```cmake
execute_process(COMMAND ${PYTHON_B} -c "import sys; print(sys.version_info.major);" OUTPUT_VARIABLE ${PYTHON}_FIND_VERSION_MAJOR OUTPUT_STRIP_TRAILING_WHITESPACE)
execute_process(COMMAND ${PYTHON_B} -c "import sys; print(sys.version_info.minor);" OUTPUT_VARIABLE ${PYTHON}_FIND_VERSION_MINOR OUTPUT_STRIP_TRAILING_WHITESPACE)
```

This is setting `${PYTHON}_FIND_VERSION_MAJOR|MINOR` according to the select python executable

nim65s:
@jcarpent I checked this PR, and I think it provides the good behavior on systems that actually have a default python version.

@gergondet is talking about issues in 20.04, which doesn't have a default python version yet, so obviously this raises a lot of issues, which I think are out of scope until the actual release of 20.04.

Anyway, with this PR, in 20.04, installing python2 & python-numpy let me configure eigenpy correctly without specifying `PYTHON_EXECUTABLE`, and installing python3 & python3-numpy let me configure eigenpy correctly by specifying `-DPYTHON_EXECUTABLE=$(which python3)`, so I think we're good :)


Unrelated: building python 2 & python 3 module is a really nice idea :D

gergondet:
@nim65s the issue on focal "preview" but it will happen if the default Python version doesn't happen to also be the latest version installed, basically when you call `find_package(Python3 ...)` the script looks for `python3.X` (for `X` in a range of known versions inside the script) then `python3` and if one happens to have the components you're looking for then it will return components related to that version although this is not your default Python 3 version (note: the problem also exists with Python2 but given that Python 2.7 is normally supposed to be the last version it doesn't matter)

Note: I don't think this situation will normally arise in Ubuntu distributions but it might for other distributions or more esoteric setup (maybe if you are attempting to build in a specific virtualenv as well)

nim65s:
@gergondet : I guess in those cases one can easily use `-DPYTHON_EXECUTABLE=/usr/bin/python3.7`, which should not be too hard for someone using an esoteric platform. At this point, we will probably face other issues with `libboost_python<insert random obscur naming scheme here>.so`, but I think we can safely wait for the issue to actually occur before trying to handle it.

jcarpent:
I will merge this PR now. We will update the script when encountering new issues.
@gergondet @nim65s Thanks for your helpful feedback.

PR #335, opened 2020-02-26
add project dependency to generated .pc, fix #320

### LINKED ISSUE #320: Convertion from pkg-config to cmake-exports without breaking changes
Hi,

In projects that are already using the generated .pc, and that we want to switch to CMake packages, it would be better to avoid breaking compatibility during the transition.

ie. if `$pkg` depends on `$lib`, and that we switch `$lib` to CMake-Exports, the old version of `$pkg` using `$lib.pc` should still work.

This means that the generated .pc should be the same (or at least implies the same behavior) as before.

For this, I think that it would be nice to populate at least the old `${_PKG_CONFIG_REQUIRES}` in the new `ADD_PROJECT_DEPENDENCY()`.

Any remarks on this ? Should I open a PR ?

nim65s:


package-config.cmake
```diff
@@ -58,6 +58,8 @@ ENDMACRO(_SETUP_PROJECT_PACKAGE_INIT)
 #
 MACRO(ADD_PROJECT_DEPENDENCY)
   list(APPEND _PACKAGE_CONFIG_DEPENDENCIES_PROJECTS "${ARGV0}")
+  string(TOLOWER "${ARGV0}" _PROJECT_DEPENDENCY)
```
gergondet: I'm not sure about this. It happens to make things work for Eigen but could break for other projects.

I think the better way to generate a working `.pc` file is to change the behaviour of `pkg_config_append_libs` so that it extracts compilation flags from the CMake target

nim65s: I'm trying to fix the `Requires:`  field of the .pc here. As far as I understand, this field needs the filename of the .pc without the suffix. Do you think we can extract this from anywhere ? Or should we just try to find this file with CMake ?

But you're right, a quick `locate .pc` shows stuff like `Qt5Core.pc`, so this lowercase is not the good solution.

gergondet: > I'm trying to fix the Requires: field of the .pc here. As far as I understand, this field needs the filename of the .pc without the suffix. Do you think we can extract this from anywhere ? Or should we just try to find this file with CMake ?

Another situation that may occur is that the find_package call and the .pc have nothing in common. For example, the recommended way to find Qt in CMake is (for example) `find_package(Qt5 COMPONENTS Core Gui Widgets)` but this does not translate to the equivalent .pc `Requires` entry which would be `Qt5Core Qt5Gui Qt5Widgets`. This also creates a problem if there's simply no equivalent `.pc` (e.g. I use `add_project_dependency(Boost COMPONENTS filesystem timer)`)

A possible solution could be to introduce an extra argument to `add_project_dependency` so that we can easily provide the equivalent `Requires` entry if applicable e.g. `add_project_dependency(Eigen3 REQUIRED PKG_CONFIG_REQUIRES eigen3)` then we consume the `PKG_CONFIG_REQUIRES` argument(s) and pass the remainder to `find_package` see https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html

but if we go that way it's probably simpler to just implement something like `pkg_config_append_requires` :D

jmirabel: > A possible solution could be to introduce an extra argument to `add_project_dependency`

I think this is a good option and is better than `pkg_config_append_requires` because it puts everything related to one package in the same place.

nim65s: The solution with `add_project_dependency(Eigen3 REQUIRED PKG_CONFIG_REQUIRES eigen3)` looks good to me, as #320 is only a transitional state, for cases where the we knew and used the .pc before, and still want it for a while, to avoid downstream breaks.

gergondet: Sounds good to me if everyone is ok with it (as long as I don't implement it :D)

Two extra things to keep in mind:
- whether the default behavior should be to consider that `PKG_CONFIG_REQUIRES` defaults to the name of the CMake package (very common case but can lead to a broken .pc) or defaults to nothing (no broken .pc but also easy to forget)
- in the first case, add an option to not append anything in Requires (or maybe it's enough that we can pass `PKG_CONFIG_REQUIRES ""`)

nim65s: I'll implement it, and would prefer the second case.

nim65s:
It looks like this is working for me. I'll make more extensive tests next week :)

nim65s:
Hi,

This works well for me, so I think this can be merged :)

gergondet:
Sorry, it seems I missed your previous comment. I'll merge right away :)

Thanks for your contribution as always

PR #348, opened 2020-03-31
dynamic-graph: set RPATH

nim65s:
Hi,

This commit fix the `wrap.so` libs generated with `DYNAMIC_GRAPH_PYTHON_MODULE`, by providing them the right `RPATH`, at least since dynamic-graph v4.

I don't understand why I have to set `BUILD_RPATH` instead of `INSTALL_RPATH`, but my tests in compiling sot-core by hand or with robotpkg both showed that the RPATH of the installed lib is not set with `ÌNSTALL_RPATH`, while it is with `BUILD_RPATH`…


jcarpent:
I think it depends on your linker and you can have a look to what I've done for Mac systems in apple.cmake file ;)

nim65s:
Maybe you are speaking of 
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/a5c65a0ee51191be3f2e2667b95830706c211bb2/apple.cmake#L27

I don't understand which directories the doc is speaking of:
> directories in the linker search path and outside the project

I'll try with that.

On a side note, in this files, you have twice
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/a5c65a0ee51191be3f2e2667b95830706c211bb2/apple.cmake#L29

One in a condition, and another out of it.

olivier-stasse:
Friendly ping

nim65s:
Time flies, sorry :/
I'll come back to this quickly

olivier-stasse:
Dear @nim65s that was just to mention that @jmirabel approved the request.

nim65s:
> I'll try with that.

I tried again with different combinations of `INSTALL_RPATH`, `BUILD_RPATH` and `INSTALL_RPATH_USE_LINK_PATH`, and only `BUILD_RPATH` gives good results, ie. `ldd $PREFIX/lib/python3.8/site-packages/dynamic_graph/sot/core/feature_point6d/wrap.so` doesn't show any `not found`, nor any path to the build dir, when ` $PREFIX/lib/dynamic-graph-plugins` was not  manually added in `$LD_LIBRARY_PATH`.

This looks like a bug in CMake, or a wrong global configuration flag somewhere, so this might not give the same results in non-linux platforms. I tested that with ArchLinux / CMake 3.18.0 & 18.04 / CMake 3.10.2

PR #351, opened 2020-04-01
Move _LIBRARIES config to Config.cmake.in

### LINKED ISSUE #350: _PKG_CONFIG_LIBS might be empty
Hi,

This line assumes that `_PKG_CONFIG_LIBS` is not empty:

https://github.com/jrl-umi3218/jrl-cmakemodules/blob/a5c65a0ee51191be3f2e2667b95830706c211bb2/package-config.cmake#L136

So, in the case of example-robot-data, this fails with:
```
CMake Error at cmake/package-config.cmake:136 (string):
  string sub-command REPLACE requires at least four arguments.
  cmake/base.cmake:155 (SETUP_PROJECT_PACKAGE_FINALIZE)
  CMakeLists.txt:9223372036854775807 (SETUP_PROJECT_FINALIZE_HOOK)
```

(nice line number, btw)

wxmerkt:
The changes to move the configuration to the `package-config.cmake` as requested in the previous PR actually broke the logic. It won't work as the macro is run at the _end of the configuration stage_. As thus, the library is not yet built and installed (surprise!), so cannot be found.

This PR moves the logic back to the config file. This was tested locally and should fix the regression introduced many weeks ago, once and for all.

It should also fix #350. This fixes the original fix for the regression in #347

Cf. https://github.com/ros/rosdistro/pull/24287#issuecomment-607300670


wxmerkt:
After this has been merged, we need yet another eigenpy release.

jcarpent:
@wxmerkt I don't think this is a good idea to populate the Config.cmake with logic. I've proposed a solution in #352 that also fixes #350. I think this is simpler.

wxmerkt:
#352 will _not_ fix the issue we are seeing on the buildfarm

wxmerkt:
@jcarpent Do you have an alternate suggestion on how to fix this or can we proceed with merging this? I'd like to get the eigenpy regression fixed today.

jcarpent:
@wxmerkt I've pushed some stuff on your branch to make it a bit more robust.

jcarpent:
There is an additional case that should be handled it is when one provides linking parameters such as `-Wl,-undefined,dynamic_lookup,`.

jcarpent:
@wxmerkt I've just pushed the robust way for checking librairies. Could you check that everything works well on your side?

wxmerkt:
@jcarpent Thank you for making the search more robust. I added one more commit with a fix to make it work: We want only the first occurrence matched to avoid the multi-arch string being found and the library hint being appended to the libraries. This now is confirmed to work.

jcarpent:
@wxmerkt Thanks a lot. This sounds great.

jmirabel:
I did not follow the discussion on this topic and had a hard time trying to understand the piece of code in Config.cmake.in that solves this issue. This code looks like we re-invented CMake...

If I understand correctly, the intent is to populate the `PKGNAME_LIBRARIES` variable from the PKG_CONFIG_LIBRARIES. Looking at the documentation of [CMakePackageConfigHelpers](https://cmake.org/cmake/help/v3.6/module/CMakePackageConfigHelpers.html), I don't get why we don't just populate `PKGNAME_LIBRARIES` with something like `@PACKAGE_CMAKE_INSTALL_PREFIX@/declare_lib_relative_path`.

This method should make the package relocatable. It won't work if the library is moved in a way which does not correspond to a package relocation.

PR #352, opened 2020-04-02
Fix packaging

### LINKED ISSUE #350: _PKG_CONFIG_LIBS might be empty

jcarpent:
It also fixes #350.


jcarpent:
Prefer #351 which seems one correct solution.

PR #353, opened 2020-04-02
Remove duplicates in pkg-config dependencies inside Config.cmake.

jmirabel:
This solves https://github.com/stack-of-tasks/roscontrol_sot/issues/19


jmirabel:
I guess no-one objects for this trivial change.

Issue #354, opened 2020-04-03
Calling FINDPYTHON in several CMakeLists.txt makes detection  fail


florent-lamiraux:
I have spent a few hours trying to understand why in a given package python was not detected properly while in other packages it was.

I finally discovered that FINDPYTHON was also called in subdirectories.

I think it would be worth detecting this error. A simple solution would be to define a variable called `FINDPYTHON_ALREADY_CALLED` in macro `FINDPYTHON` in python.cmake and to check that this variable has not been not already defined when calling `FINDPYTHON`.

jcarpent:
Solved by #355.

PR #355, opened 2020-04-03
[python.cmake] Detect multiple calls to FINDPYTHON

florent-lamiraux:
Raise FATAL_ERROR is FINDPYTHON is called several times in the same project.
This solves issue https://github.com/jrl-umi3218/jrl-cmakemodules/issues/354.


nim65s:
This seems a bit excessive. Could we just not do anything, if we detect that this macro has already been called before ?

nim65s:
`FINDPYTHON` is already been called in different places in this module. I think most of them are protected by checking that it has not already been called, but we would have to enforce that.

But I think calling it should not be "FATAL", as this is not by itself a reason to prevent anything to configure or compile. A big annoying AUTHOR_WARNING should do, but we can wait for others to comment :)

jcarpent:
The main question is why two calls of FINDPYTHON are resulting in two different outputs?

florent-lamiraux:
@jcarpent : I do not know.
@nim65s : I tried with ERROR instead of FATAL_ERROR, but I did not find cmake output very expressive. cmake went on configuring and complained that python was not detected correctly.
If we agree on my solution (but I also think we should wait for other people to react), I will add a comment in the documentation of the macro to specify that it should not be called several times.

jmirabel:
FIND_PYTHON may be called several times in metabuild systems like catkin. I agree that we should rather investigate why two calls gives different output. in the meantime, I think we should only add a user warning that the macro was called several times.

@florent-lamiraux were the arguments the same for all the calls ?

jmirabel:
> I think we should only add a user warning that the macro was called several times.

Not a user warning but rather an AUTHOR_WARNING.

florent-lamiraux:
> @florent-lamiraux were the arguments the same for all the calls ?
There was no argument in each call.

florent-lamiraux:
I have just replaced FATAL_ERROR by AUTHOR_WARNING.

python.cmake
```diff
@@ -54,7 +54,10 @@ IF(CMAKE_VERSION VERSION_LESS "3.2")
 ENDIF(CMAKE_VERSION VERSION_LESS "3.2")
 
 MACRO(FINDPYTHON)
-
+  IF(DEFINED FINDPYTHON_ALREADY_CALLED)
+    MESSAGE(AUTHOR_WARNING "Macro FINDPYTHON has already been called.")
```
jmirabel, line 58:
```suggestion
    MESSAGE(AUTHOR_WARNING "Macro FINDPYTHON has already been called. Several call to FINDPYTHON may not find the same Python version (for a yet unknown reason).")
```

nim65s:
Until a better solution is found, this is also fine for me.

But to fix the issue, I suggest to just return from the macro after the AUTHOR_WARNING.

But we can't return from macros in CMake, as they are expanded in place. Could we use a function ? https://cmake.org/cmake/help/latest/command/return.html

jmirabel:
> But to fix the issue, I suggest to just return from the macro after the AUTHOR_WARNING.

This fixes the symptoms, not the issue and it adds one more *IF* to the logic.

florent-lamiraux:
The difference between the first and second calls seems to be in `FIND_PROGRAM(PYTHON_EXECUTABLE python)` in `/usr/share/cmake-3.10/Modules/FindPythonInterp.cmake`.
I do not know how to debug this.

florent-lamiraux:
Fine. I take @jmirabel 's suggestion and merge.
Is it okay for everyone ?

PR #356, opened 2020-04-03
Strip and remove duplicates in pkg-config dependencies inside Config.cmake.

jmirabel:


jmirabel:
This completes #353

jcarpent:
Please wait, discussions are occuring here #350.

jcarpent:
Better to fix #350 than providing pacthes which will be removed then.

jcarpent:
In fact, I did see that you were not changing the same thing

jmirabel:
Thanks for your quick reviews.

Issue #341, opened 2020-03-12
Doxygen warnings


wxmerkt:
During compilation, I get a set of warnings related to the `Doxyfile`:
```
warning: ignoring unsupported tag `OUTPUT_TEXT_DIRECTION  =' at line 90, file Doxyfile
warning: ignoring unsupported tag `JAVADOC_BANNER         =' at line 176, file Doxyfile
warning: ignoring unsupported tag `OPTIMIZE_OUTPUT_SLICE  =' at line 257, file Doxyfile
warning: ignoring unsupported tag `EXTRACT_PRIV_VIRTUAL   =' at line 409, file Doxyfile
warning: ignoring unsupported tag `CLANG_DATABASE_PATH    =' at line 932, file Doxyfile
warning: ignoring unsupported tag `HTML_DYNAMIC_MENUS     =' at line 1061, file Doxyfile
warning: ignoring unsupported tag `FORMULA_MACROFILE      =' at line 1291, file Doxyfile
warning: ignoring unsupported tag `LATEX_MAKEINDEX_CMD    =' at line 1448, file Doxyfile
warning: ignoring unsupported tag `LATEX_EMOJI_DIRECTORY  =' at line 1552, file Doxyfile
warning: ignoring unsupported tag `XML_NS_MEMB_FILE_SCOPE =' at line 1665, file Doxyfile
```

This is with doxygen 1.8.13 on Ubuntu 18.04. I am happy to make a PR to remove these options from the Doxyfile unless you see a need to keep them?

gergondet:
I think these tags are supported in later versions of Doxygen. This was introduced by #314 and I'm guessing @aescande has a recent version of Doxygen.

I don't know whether there is a way to filter out unsupported based on Doxygen version. Since the number of new options is relatively small we could also filter them manually based on Doxygen version

jmirabel:
Another option to do this:
1. use `doxygen -g doxyfile_just_for_checking` to generate a config file.
2. check if option is present in doxyfile_just_for_checking before adding it to Doxyfile.

florent-lamiraux:
I am reluctant about detecting automatically which variables are supported. It will add a source of bug.
I would suggest in the future to add new variables in the Doxifile only if someones needs them. Let us recall that `Doxifile.extra.in` can be used to add variables per project.

jcarpent:
Solved by #360.

PR #358, opened 2020-04-04
[Doxigen] Remove obsolete variables.

florent-lamiraux:


gergondet:
I don't think those are obsolete, rather the opposite, see #341

florent-lamiraux:
You are probably right. Then we can leave it like this and wait for those warnings to disappear.
In the future, however, I would suggest to add new variables to `Doxyfile.in` only if someone needs it.
Let us recall that people can add variables in `Doxifile.extra.in`

PR #367, opened 2020-04-10
[HPP] Fix layout file.

jmirabel:


jmirabel:
This PR also fixes the default value that changed due to https://github.com/jrl-umi3218/jrl-cmakemodules/pull/360

PR #375, opened 2020-04-30
Do not duplicate pkg-config dependencies.

jmirabel:
This fixes:
- https://github.com/humanoid-path-planner/hpp-manipulation-corba/issues/92
- https://github.com/jrl-umi3218/jrl-cmakemodules/pull/356
- https://github.com/stack-of-tasks/roscontrol_sot/issues/19


jmirabel:
I was over-optimistic. The duplication is less important but still a way too big. Now the doxytag string is 15000 (instead of 30000) in hpp-manipulation-corba package.

jmirabel:
After some digging, it turns out I can remove the dependency list by modifying the CMakeLists.txt of a package dependency (removing repeated calls to find_package on the same package).

Clearly this is not a solution. The two last commits provide some ways of breaking the recursion when relevant. They help, but I don't understand why they don't completely remove duplicated calls.

Any comment on the two methods is welcome.

jmirabel:
Any comments on this ?

jmirabel:
I pushed the code into branch `avoid_duplicated_dependency_lookup` so that I can move on.

jmirabel:
@florent-lamiraux the last commit should fix the trailing ',' issue.

florent-lamiraux:
I will update my packages and retry.

nim65s:
This is breaking:
- all builds in 16.04, with `CMake Error at cmake/pkg-config.cmake:24 (list): list does not recognize sub-command JOIN` (the guard on CMake version doesn't seem to work…)
- some projects (eg. sot-core), where quotes are missing, leading to `CMake Error at cmake/pkg-config.cmake:24 (list): list sub-command JOIN requires three arguments (10 found).`

PR #377, opened 2020-05-08
Fix version check + usage of list(JOIN)

jmirabel:
Fixes https://github.com/jrl-umi3218/jrl-cmakemodules/pull/375#issuecomment-625563903


nim65s:
I can confirm that this fixed all the issues I had :)
Thanks !

PR #387, opened 2020-05-25
[python.cmake] Uses UNIX links only with UNIX.

olivier-stasse:
In DYNAMIC_GRAPH_PYTHON_MODULE when trying to compile dynamic-python-graph on MacOS X gcc provided by the homebrew environment is not able to understand the linking option no-as-needed.

This patch fixes the issue.


olivier-stasse:
Please do not merge

olivier-stasse:
Sorry for the delay. I think this is fine now.

olivier-stasse:
Thanks @gergondet !

PR #389, opened 2020-05-26
[win32] Handle absolute paths for python generation.

francois-keith:
If the path to the python file to compile is absolute,
remplace the : at the beginning of the path by _
e.g. ``C:/tmp/file.py`` becomes ``C__tmp_file.py``


jmirabel:
I think the compile file is just garbage and is not used anymore (used to be installed) so this is reasonable.

jmirabel:
On windows, cannot you have paths with `\` too ?

francois-keith:
Normally, yes, but the regex is applied on the string "${CMAKE_CURRENT_BINARY_DIR}/${MODULE}/${FILE}c" that does not seem to contain the '\' character (this is a cmake-generated variable).

PR #392, opened 2020-05-26
Fix windows issue on target name

jcarpent:


python.cmake
```diff
@@ -394,6 +394,12 @@ MACRO(PYTHON_BUILD MODULE FILE)
   STRING(REGEX REPLACE "/|:" "_" FILE_TARGET_NAME
     "${CMAKE_CURRENT_BINARY_DIR}/${MODULE}/${FILE}c")
 
+  # Remove extra C:\* on Windows
+  IF(WIN32)
+    STRING(REGEX REPLACE "^[^:_]*:_" "" FILE_TARGET_NAME
+      "${FILE_TARGET_NAME}")
+  ENDIF(WIN32)
+
```
jmirabel, line 402: If this issue is too troublesome, we can use add_custom_command instead.

jcarpent: I agree. This would be the cleanest solution.

jcarpent: @nim65s What do you think?

nim65s: I don't have any useful comment for code enclosed in `IF(WIN32)`.

jcarpent: I was mentioning the suggestion made by @jmirabel to gather together all the generated Python files.

jcarpent:
I've just updated the content of this PR and followed the recommandation made by @jmirabel which is very smooth.

python.cmake
nim65s: Now I get it. Yes, that was clearly a good idea :)

PR #394, opened 2020-05-29
Add missing config when creating Python test

jcarpent:


jcarpent:
This feature is requested on Windows for instance

jmirabel:
Did you check this works find on Linux too ?

nim65s:
According to the [CMake 3.1 doc](https://cmake.org/cmake/help/v3.1/command/add_test.html), 

> The COMMAND and WORKING_DIRECTORY options may use “generator expressions” with the syntax $<...>. See the cmake-generator-expressions(7) manual for available expressions.

So the generator expression for `CONFIGURATION` doesn't seem available.

jcarpent:
This is not the right to operate. I will close this PR then.

PR #395, opened 2020-06-05
Enhance WIN32 compatibility

jcarpent:


test.cmake
```diff
@@ -108,6 +108,9 @@ MACRO(ADD_PYTHON_UNIT_TEST NAME SOURCE)
   SET(MODULES "${ARGN}")  # ARGN is not a variable
   FOREACH(MODULE_PATH IN LISTS MODULES)
     LIST(APPEND PYTHONPATH "${CMAKE_BINARY_DIR}/${MODULE_PATH}")
+    IF(WIN32)
+      LIST(APPEND PYTHONPATH "${CMAKE_BINARY_DIR}/${MODULE_PATH}/$<CONFIG>")
+    ENDIF(WIN32)
```
jmirabel, line 113: shouldn't there be a `else` here ?

jcarpent: Not in the way we are generating the Python files currently.

jcarpent:
OK. I can do that. My guess was my change are quoted by some `IF(WIN32)` chaning nothing for the current users.

Issue #398, opened 2020-06-09
Exporting <package-name>Config.cmake files with no targets.


olivier-stasse:
When having pure data package, typically robot_description kind of packages, the <package-name>Config.cmake files are very handy.
However they do not have targets. In the current writing of the SETUP_PROJECT_PACKAGE_FINALIZE macro the INSTALL( export ) calls fails because they are no targets to export.
IMHO the correct behavior is to not call this install when there is no target.

nim65s:
Hi @olivier-stasse,

Where do you have this issue ?
All SoT packages have an  `INTERFACE` target,  eg. :

https://github.com/stack-of-tasks/tiago_data/blob/75750a1da675319f30534e3383a657e9178e8d60/CMakeLists.txt#L19

olivier-stasse:
talos_data when gazebo is not detected.

nim65s:
Can you please open an issue on talos-data, before working further on that topic ? This looks like https://github.com/stack-of-tasks/talos-data/pull/2 to me.

olivier-stasse:
I respectfully disagree. jrl-cmakemodules should be able to handle package with no target.

olivier-stasse:
But of course this is a personal opinion that should be discussed openly and I will go towards  the community decision.
Note: my solution is partial. Because Config.cmake.in impose to include Targets.cmake this later trigger an error on a package using it.

nim65s:
I agree, jrl-cmakemodules should be able to handle package with no target. But I would like to understand the issue you are trying to solve on talos-data.

olivier-stasse:
Fixed by PR #408

PR #399, opened 2020-06-09
Fix isssue #398

olivier-stasse:
Do not call INSTALL(EXPORT ...) when having no targets but generates the Config.cmake file.

### DIFF
--- a/package-config.cmake
+++ b/package-config.cmake
@@ -168,9 +168,12 @@ install(
 
 # Config
 #   * <prefix>/lib/cmake/Foo/FooTargets.cmake
-install(
+if(${TARGETS_EXPORT_NAME})
+  install(
     EXPORT "${TARGETS_EXPORT_NAME}"
     NAMESPACE "${namespace}"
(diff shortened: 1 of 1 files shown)


gergondet:
This looks nice. Did you try this patch on such a package?

Note that for description package that are handled by catkin this might cause conflict with the Config file generated/installed by catkin.

olivier-stasse:
> This looks nice. 
Thanks

> Did you try this patch on such a package?
Yes, and it does not work without this modification.
 
> Note that for description package that are handled by catkin this might cause conflict with the Config file generated/installed by catkin.
Indeed but the conflict arises only if you use catkin_package
cf https://github.com/ros/catkin/blob/noetic-devel/cmake/catkin_package.cmake
For some reason INSTALL(EXPORT ) is not used by catkin_package.

Note that our strategy of using the hook does not let the user any choice.

gergondet:
I'm reverting this, it breaks packages that do export targets.

Issue #391, opened 2020-05-26
Exporting Boost components


jcarpent:
A small remark to mention that at the time being, the Boost components are not exported at all by a project. We need to take on the fact that some of them are only BUILD dependencies, and should not be exported then.

gergondet:
You can/should use `add_project_dependency` with projects dependencies and `find_package` with build time dependencies

So for example:
```
add_project_dependency(Boost REQUIRED COMPONENTS thread)
target_link_libraries(myLib PUBLIC Boost::thread)
# find_package(MyPackage) will also find_package(Boost REQUIRED COMPONENTS thread)

find_package(Boost REQUIRED COMPONENTS test)
target_link_libraries(myTest PUBLIC Boost::test myLib) # still links to Boost::thread
```

Note: if you only need Boost headers (i.e. the `Boost::boost` target) you can omit the `COMPONENTS` if you require `cmake > 3.5` (notably, Ubuntu 16.04 has CMake 3.5)

Note 2: I'm not sure how well this works with boost-python as I don't use it

nim65s:
Do you mean `cmake >= 3.5` ?
If so, I think we can require CMake 3.5.

Debian 8 (oldoldstable) has CMake 3.0, so it's broken anyway (ref #338), and Debian >= 9 (oldstable, 2017) has 3.7. Fedora 27 (2017) has 3.11.

gergondet:
Nope, I really mean `cmake > 3.5` 

I'm basing this on [a comment I made in a CMakeLists.txt](https://github.com/jrl-umi3218/mc_rbdyn_urdf/commit/7ac0a1e3f7d9acb7b1221fad32fd5538c7abe9c5) which I'm pretty sure I added after a build failure on Xenial (I was still building on Trusty at the time but with CMake 3.5 too)

Note: this can also be worked-around by using a more recent `FindBoost.cmake`

nim65s:
Ok, thanks

jcarpent:
So, currently in the module, we have a specific macro to handle Boost components. It would be nice to add the features of Optional and Required and Build components. In general, Boost.Python is not exported to link the target to an external target.
@nim65s Would have time to handle it? Do you see any good solution?

nim65s:
I guess this is just a matter of adding parameters to the macro, and using
```cmake
add_project_dependency(Boost REQUIRED COMPONENTS ${BOOST_REQUIRED_COMPONENTS} REQUIRED)
[...]

PR #402, opened 2020-06-13
[package-config.cmake] Handle in a friendly user manner optional package

olivier-stasse:
An optional package is specified in the following cases:
ADD_PROJECT_DEPENDENCY(package)
ADD_PROJECT_DEPENDENCY(package OPTIONAL)
In this case if the package is not detected and this is not put in the dependency list.


nim65s:
Hi,

`ADD_PROJECT_DEPENDENCY(package)` is already optional ; because if it is not, you should use
`ADD_PROJECT_DEPENDENCY(package REQUIRED)`.

Moreover, this macro mimics the API of the upstream ones it calls: `find_package` / `find_dependency`, so I think we should try to stick to this API as much as possible, unless there an issue you want to solve ?

jmirabel:
1. is it an issue to have an `find_package(Foo)` in the BarConfig.cmake ? Even if, for aliasing reason, Foo wasn't found at compile time but is found by packages depending on Bar, the targets from Foo will not be used by Bar.
2. to avoid adding optional dependencies to list of dependencies, you should merely add them after the `find_package`, only if `package_FOUND` is true. It will cover all cases, with no change in the API and a very minimal change in the code.

olivier-stasse:
* I am fine with not putting OPTIONAL (i just find it more clear to the user)
 * If I am not mistaken right now, I have to do:
```
find_package(gazebo)
if (GAZEBO_FOUND)
  ADD_PROJECT_DEPENDENCY(gazebo)
endif(GAZEBO_FOUND)
```
the proposed PR simply replace those lines by:
```
ADD_PROJECT_DEPENDENCY(gazebo)
```
if REQUIRED is not specified then it is optional.
Does that make sense ?

jmirabel:
>     * If I am not mistaken right now, I have to do:
> 
> 
> ```
> find_package(gazebo)
> if (GAZEBO_FOUND)
>   ADD_PROJECT_DEPENDENCY(gazebo)
> endif(GAZEBO_FOUND)
> ```

Why do you have to do this ?

olivier-stasse:
As far as I tested ADD_PROJECT_DEPENDENCY will add Gazebo to the list of dependency even if it is not present.
So you need to quietly test it before, and then add it if it is detected.
See https://github.com/jrl-umi3218/jrl-cmakemodules/issues/391 on how Pierre is handling optional package.

package-config.cmake
```diff
@@ -70,21 +70,49 @@ MACRO(ADD_PROJECT_DEPENDENCY)
...
-  string(REPLACE ";" " " PACKAGE_ARGS "${PARSED_ARGN_UNPARSED_ARGUMENTS}")
-  list(APPEND _PACKAGE_CONFIG_DEPENDENCIES_FIND_PACKAGE "  find_package(${PACKAGE_ARGS})")
-  list(APPEND _PACKAGE_CONFIG_DEPENDENCIES_FIND_DEPENDENCY "  find_dependency(${PACKAGE_ARGS})")
-  find_package(${PARSED_ARGN_UNPARSED_ARGUMENTS})
+  string(REPLACE ";" " " PACKAGE_ARGS
+    "${PARSED_ARGN_UNPARSED_ARGUMENTS}")
+
+  # If REQUIRED is not specified consider the package as optional
+  # search for the package in a quiet mode.
+  IF(PARSED_ARGN_OPTIONAL OR NOT PARSED_ARGN_REQUIRED)
```
gergondet, line 92: Please avoid:
- using all caps IF/ELSE/ENDIF
- using the matching arguments in else and endif conditions, they have not been required for a really long time

package-config.cmake
```diff
@@ -70,21 +70,49 @@ MACRO(ADD_PROJECT_DEPENDENCY)
...
+    find_package(${PARSED_ARGN_UNPARSED_ARGUMENTS} REQUIRED)
+  ENDIF(PARSED_ARGN_OPTIONAL OR NOT PARSED_ARGN_REQUIRED)
+
+  # If the package has been found provides the dependency
+  IF(${${ARGV0}_FOUND})
+
+    list(APPEND _PACKAGE_CONFIG_DEPENDENCIES_FIND_PACKAGE
+      "  find_package(${PACKAGE_ARGS})")
+    list(APPEND _PACKAGE_CONFIG_DEPENDENCIES_FIND_DEPENDENCY
+      "  find_dependency(${PACKAGE_ARGS})")
```
gergondet, line 105: The generated file will miss the `REQUIRED` argument since you consume it in `cmake_parse_arguments`

jmirabel:
First of all, the issue you mention is purely cosmetic. The only difference between
```cmake
find_package(gazebo)
if (GAZEBO_FOUND)
  ADD_PROJECT_DEPENDENCY(gazebo)
endif(GAZEBO_FOUND)
```
and
```cmake
ADD_PROJECT_DEPENDENCY(gazebo)
```
is that, in the second case, the `Config.cmake` will look for gazebo (without failing if not found) and will never use this dependency.

jmirabel:
So both version should have the same behavior.

Now, if you really want to solve the cosmetic issue, I think the API does not have to be modified. You can simply do
```cmake

MACRO(ADD_PROJECT_DEPENDENCY)
  # add dependency to the generated .pc
  # ref https://github.com/jrl-umi3218/jrl-cmakemodules/pull/335
  cmake_parse_arguments(PARSED_ARGN "" "PKG_CONFIG_REQUIRES" "" ${ARGN})
  IF(PARSED_ARGN_PKG_CONFIG_REQUIRES)
    _ADD_TO_LIST_IF_NOT_PRESENT(_PKG_CONFIG_REQUIRES "${PARSED_ARGN_PKG_CONFIG_REQUIRES}")
    _ADD_TO_LIST_IF_NOT_PRESENT(_PKG_CONFIG_DEP_NOT_FOR_CONFIG_CMAKE "${PARSED_ARGN_PKG_CONFIG_REQUIRES}")
  ENDIF()
  
  find_package(${PARSED_ARGN_UNPARSED_ARGUMENTS})
  if (${ARGV0}}_FOUND)

    list(APPEND _PACKAGE_CONFIG_DEPENDENCIES_PROJECTS "${ARGV0}")


    string(REPLACE ";" " " PACKAGE_ARGS "${PARSED_ARGN_UNPARSED_ARGUMENTS}")
    list(APPEND _PACKAGE_CONFIG_DEPENDENCIES_FIND_PACKAGE "  find_package(${PACKAGE_ARGS})")
    list(APPEND _PACKAGE_CONFIG_DEPENDENCIES_FIND_DEPENDENCY "  find_dependency(${PACKAGE_ARGS})")

    # Propagate variables changes to the cached values
    set(_PACKAGE_CONFIG_DEPENDENCIES_PROJECTS "${_PACKAGE_CONFIG_DEPENDENCIES_PROJECTS}" CACHE INTERNAL "")
    set(_PACKAGE_CONFIG_DEPENDENCIES_FIND_PACKAGE "${_PACKAGE_CONFIG_DEPENDENCIES_FIND_PACKAGE}" CACHE INTERNAL "")
    set(_PACKAGE_CONFIG_DEPENDENCIES_FIND_DEPENDENCY "${_PACKAGE_CONFIG_DEPENDENCIES_FIND_DEPENDENCY}" CACHE INTERNAL "")

  endif()
ENDMACRO()
```

olivier-stasse:
Thanks for the comments, I will stick to the solution suggested by @gergondet for the optional part in #391

package-config.cmake
olivier-stasse: Thanks @gergondet for the feedback. I wanted to update my skills in cmake. 
For the sake of my understanding: 

* if the macro is using caps should we switch to lower case or stay coherent with the macros.

* What is the proper way to deal with arguments in a macro without if/else/endif conditions ? Could you point me towards an example ?

gergondet: > if the macro is using caps should we switch to lower case or stay coherent with the macros.

I think within a single macro we should stay coherent (or use the opportunity to switch all the macro to lower-cases)

Note: I realize it's not the case in this macro but I didn't review that part myself (and we could use some linter/format integration in the process)

> What is the proper way to deal with arguments in a macro without if/else/endif conditions ? Could you point me towards an example ?

In that case, I meant to replace:

```cmake
if(PARSED_ARGN_OPTIONAL OR NOT PARSED_ARGN_REQUIRED)
...
else(PARSED_ARGN_OPTIONAL OR NOT PARSED_ARGN_REQUIRED)
...
endif(PARSED_ARGN_OPTIONAL OR NOT PARSED_ARGN_REQUIRED)
```

with

```cmake
if(PARSED_ARGN_OPTIONAL OR NOT PARSED_ARGN_REQUIRED)
...
else()
...
endif()
```

the repeated conditions have not been necessary for a long time (CMake 2.6.0 see [this answer](https://stackoverflow.com/a/35269674))

package-config.cmake
olivier-stasse: Yes we would need to test PARSED_ARGN_REQUIRED to add or not the REQUIRED parameter.

package-config.cmake
olivier-stasse: Thanks for the clarification !

PR #407, opened 2020-06-17
Make Boost components exportable.

jmirabel:


jmirabel:
I tried it in hpp-fcl and it worked fine. Could we get this merge ?

nim65s:
Do we have a way to specify components that don't need to be exported, eg. unit tests ?

jmirabel:
For those ones, you can use `find_packge` directly.

jmirabel:
Actually, the only reason to keep `SEARCH_FOR_BOOST` is Boost Python. We could remove altogether SEARCH_FOR_BOOST and keep only a single SEARCH_FOR_BOOST_PYTHON macro.

PR #410, opened 2020-06-23
[Python] Allow user to provide PYTHON_SITELIB

nim65s:
Hi,

This is just englobing the `PYTHON_SITELIB` computation in a guard, allowing the user to directly provide it. This was requested in https://github.com/Gepetto/example-robot-data/pull/38.

eg.
```bash
╰─>$ cmake -DCMAKE_INSTALL_PREFIX=/tmp/foo -DPYTHON_SITELIB=/tmp/bar ..
[…]
-- Found Python3: /usr/bin/python (found version "3.8.3") found components: Interpreter Development 
-- PythonLibraryDirs: /usr/lib
-- PythonLibVersionString: 3.8.3
-- Python site lib: /tmp/bar
```

(obviously, the site lib can also be inside the install prefix)

### DIFF
--- a/python.cmake
+++ b/python.cmake
@@ -80,7 +80,7 @@ MACRO(FINDPYTHON)
         OUTPUT_STRIP_TRAILING_WHITESPACE
         ERROR_STRIP_TRAILING_WHITESPACE
         )
-      
+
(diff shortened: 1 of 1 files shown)


nim65s:
As usual, [this is easier to read without looking at spaces](https://github.com/jrl-umi3218/jrl-cmakemodules/pull/410/files?w=1)

Issue #413, opened 2020-07-13
Boost 1.72


olivier-stasse:
I am currently having the following problem on MacOS Catalina.
In sot-core, jrl-cmakemodules is detecting the following:
```sh
Boost_PYTHON_LIBRARY=optimized;/usr/local/lib/libboost_python38.dylib;debug;/usr/local/lib/libboost_python38-mt.dylib
```
which is then used by ```Boost_PYTHON_LIBRARIES```
here https://github.com/jrl-umi3218/jrl-cmakemodules/blob/c333a88decb3e4c0a86947bc6c7f072dc5c5df20/boost.cmake#L227

It thus creates the following option at link time
```sh
-Wl,-undefined,dynamic_lookup,optimized
```
as far as I understood optimized should be replaced by the second argument of ```Boost_PYTHON_LIBRARY```, or maybe by ```Boost_PYTHON_LIBRARY_RELEASE```if this is the mode set for ```CMAKE_BUILD_TYPE```.

Did somebody else had the problem ?

jcarpent:
@olivier-stasse Do you still encounter this issue?

jcarpent:
It is not appearing on my side.

olivier-stasse:
I still have the same issue.

olivier-stasse:
Ok. It seems that the branch topic/dynamic_graph_cp11 is missing some macros namely SET_BOOST_DEFAULT_OPTIONS in boost.cmake which, once included, provides the expected behavior.
So I guess I just have to wait that this branch gets properly merged in the master branch.

nim65s:
I merged master into that branch this morning

olivier-stasse:
I have seen that afterwards.
Many thanks @nim65s

PR #414, opened 2020-07-13
[boost] Replace specific call to the linker by target Boost::python.

olivier-stasse:
Fix for #413
Replace explicit  ```-Wl,-undefined,dynamic_lookup,${Boost_PYTHON_LIBRARIES} ``` by ```Boost::python```


boost.cmake
```diff
@@ -224,7 +224,7 @@ MACRO(TARGET_LINK_BOOST_PYTHON target)
     IF(${TARGET_TYPE} MATCHES EXECUTABLE)
       TARGET_LINK_LIBRARIES(${target} ${PUBLIC_KEYWORD} ${Boost_PYTHON_LIBRARY})
     ELSE(${TARGET_TYPE} MATCHES EXECUTABLE)
-      TARGET_LINK_LIBRARIES(${target} ${PUBLIC_KEYWORD} -Wl,-undefined,dynamic_lookup,${Boost_PYTHON_LIBRARIES})
```
jcarpent, line 227: This won't work for old version of Boost where no target are exported.

PR #417, opened 2020-07-17
Fix warning

jcarpent:
To avoid the warning as refered in http://build.ros.org/job/Kdev__eigenpy__ubuntu_xenial_amd64/46/cmake/new/folder.94772433/category.2333/


jcarpent:
@jmirabel Can you check this PR?

jmirabel:
I believe the correct fix is ed1152c

jcarpent:
OK. Could you push on top of this?

jmirabel:
I don't think it is a good idea to introduced old policies so it would be better to remove your commit. Don't you think ?

jcarpent:
Yep, I do agree. I close this PR and you open a new one?

PR #418, opened 2020-07-17
Fix usage of _ADD_TO_LIST_IF_NOT_PRESENT

jmirabel:


jmirabel:
Related to #417.

jcarpent:
The error is still there:
```
CMake Warning (dev) at cmake/base.cmake:203 (IF):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "_PACKAGE_CONFIG_DEPENDENCIES_PROJECTS" will no
  longer be dereferenced when the policy is set to NEW.  Since the policy is
  not set the OLD behavior will be used.
  cmake/package-config.cmake:79 (_ADD_TO_LIST_IF_NOT_PRESENT)
  CMakeLists.txt:67 (ADD_PROJECT_DEPENDENCY)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

jmirabel:
On eigenpy, the devel and master branch uses 1498ee6. Did you update the git submodule ?

package-config.cmake
```diff
@@ -76,11 +76,11 @@ MACRO(ADD_PROJECT_DEPENDENCY)
   if(PARSED_ARGN_FOR_COMPONENT)
     set(component "_${PARSED_ARGN_FOR_COMPONENT}")
   endif(PARSED_ARGN_FOR_COMPONENT)
-  _ADD_TO_LIST_IF_NOT_PRESENT(APPEND "_PACKAGE_CONFIG${component}_DEPENDENCIES_PROJECTS" "${ARGV0}")
+  _ADD_TO_LIST_IF_NOT_PRESENT(APPEND _PACKAGE_CONFIG${component}_DEPENDENCIES_PROJECTS "${ARGV0}")
```
jmirabel, line 79: I think the mistake is here. `APPEND` should not be here...

PR #420, opened 2020-07-20
Fix variable name

jcarpent:


jcarpent:
According to https://cmake.org/cmake/help/v3.0/module/FindBoost.html

PR #421, opened 2020-07-20
Export Boost search options

jcarpent:


jcarpent:
This is required to find first the multithreaded version + the shared lib version

Issue #422, opened 2020-07-28
Boost::Python export on 16.04


nim65s:
Hi,

Following #407, I have several errors about the fact that we export a Boost::Python dependency under a made up name.

Here is the first.

On 16.04 (CMake 3.5), I can build and install eigenpy fine with Python 3. But then, when I want to configure hpp-fcl:
```
-- eigenpy searched on  FOUND. eigenpy at /tmp/3/lib/libeigenpy.so
-- boost_python-py35 searched on  FOUND. boost_python-py35 at /usr/lib/x86_64-linux-gnu/libboost_python-py35.so
CMake Error at /usr/share/cmake-3.5/Modules/FindBoost.cmake:1677 (message):
  Unable to find the requested Boost libraries.

  Boost version: 1.58.0

  Boost include path: /usr/include

  Could not find the following Boost libraries:

          boost_python35

  No Boost libraries were found.  You may need to set BOOST_LIBRARYDIR to the
  directory containing Boost libraries or BOOST_ROOT to the location of
  Boost.
Call Stack (most recent call first):
  /tmp/3/lib/cmake/eigenpy/eigenpyConfig.cmake:123 (find_package)
  CMakeLists.txt:75 (FIND_PACKAGE)
```

Where `CMakeLists.txt:75` is `FIND_PACKAGE(eigenpy 2.2 REQUIRED)`, and `/tmp/3/lib/cmake/eigenpy/eigenpyConfig.cmake:123` is `find_package(Boost COMPONENTS python35 REQUIRED)`.

This issue is [breaking build on robotpkg](http://robotpkg.openrobots.org/rbulk/robotpkg/path/py-hpp-fcl/py35-hpp-fcl-1.4.5/Ubuntu-16.04.pal-x86_64/bulk.html)

Adding the generated `find_package(Boost COMPONENTS python35 REQUIRED)` in various places, I found that the issue is raised if this line is added before that one

https://github.com/jrl-umi3218/jrl-cmakemodules/blob/fb4c22c319ec5320f9a85527eb1a4130954846f5/boost.cmake#L61

So, should we ship this `FindBoost.cmake` ? Or do we have another solution ?
Removing Boost::Python from the exports would be a workaround too.

In any case, I think that we should ensure that the ways we find Boost and its components are the same, whether we are in the main CMakeLists.txt or in a generated xxxConfig.cmake.

jcarpent:
[...]

PR #424, opened 2020-07-28
Add Boost_NO_BOOST_CMAKE to generated xxxConfig.cmake.

### LINKED ISSUE #423: Boost::Python export on Arch
Hi,

Following #407, I have several errors about the fact that we export a Boost::Python dependency under a made up name.

Here is the second, following #422

On ArchLinux (CMake 3.18), I get:
```
-- eigenpy searched on  FOUND. eigenpy at /tmp/3/lib/libeigenpy.so
-- boost_python3 searched on  FOUND. boost_python3 at /usr/lib/libboost_python3.so
CMake Error at /usr/lib64/cmake/Boost-1.72.0/BoostConfig.cmake:120 (find_package):
  Could not find a package configuration file provided by "boost_python3"
  (requested version 1.72.0) with any of the following names:

    boost_python3Config.cmake
    boost_python3-config.cmake

  Add the installation prefix of "boost_python3" to CMAKE_PREFIX_PATH or set
  "boost_python3_DIR" to a directory containing one of the above files.  If
  "boost_python3" provides a separate development package or SDK, be sure it
  has been installed.
  /usr/lib64/cmake/Boost-1.72.0/BoostConfig.cmake:185 (boost_find_component)
  /usr/share/cmake-3.18/Modules/FindBoost.cmake:448 (find_package)
  /usr/share/cmake-3.18/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  /tmp/3/lib/cmake/eigenpy/eigenpyConfig.cmake:128 (find_dependency)
  CMakeLists.txt:75 (FIND_PACKAGE)
```

I'll open a PR for this.

nim65s:
Hi,

This PR fix #423, by applying the strategy from #421.


nim65s:
On a second though, we should probably add the whole test for 1.70 in the generated config file, no matter with which version of Boost was detected when this config file was generated.

jmirabel:
Can you add the rationale for set this variable as a comment into the file ? It seems very unnatural to me...

nim65s:
Well, actually, this is just to mimick https://github.com/jrl-umi3218/jrl-cmakemodules/blob/fb4c22c319ec5320f9a85527eb1a4130954846f5/boost.cmake#L74 that came from #216. 

I hope at some point, we will be able to use those `Boost::Component` things, but it looks like this is not there yet, as said in https://github.com/jrl-umi3218/jrl-cmakemodules/pull/414#discussion_r455192804

Anyway, I still don't have good ideas to fix #422, so maybe the easiest workaround for now is to stop exporting boost python.

jmirabel:
Doesn't Boost::Component work expect for Boost Python ? At least, I think it is the case on ubuntu 18.04.

nim65s:
Closed in favor of #430

PR #426, opened 2020-07-31
[Python] allow compile_pyc from different directories, fix #425

### LINKED ISSUE #425: TARGET 'compile_pyc' was not created in this directory.
Hi,

As noted by @proyan in https://github.com/loco-3d/crocoddyl/pull/809#issuecomment-664992804, #392 introduced a custom target, which break this macro when it is used from multiple CMakeLists.txt.

nim65s:
Hi,

This might not be the best workaround, but I think this does the job.

I'm not sure what was the exact issue on windows that #392 had to fix, though.

### DIFF
--- a/python.cmake
+++ b/python.cmake
@@ -398,13 +398,15 @@ ENDMACRO()
 # Build a Python file from the source directory in the build directory.
 #
 MACRO(PYTHON_BUILD MODULE FILE)
-  IF(NOT TARGET compile_pyc)
-    ADD_CUSTOM_TARGET(compile_pyc ALL)
+  # Regex from IsValidTargetName in CMake/Source/cmGeneratorExpression.cxx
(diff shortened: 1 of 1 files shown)


jmirabel:
I think the issue is that the characters allowed for target names are not the same as the one for filenames.

nim65s:
Ok, then this PR ensure that we only have a target name with allowed characters for target names.

nim65s:
I don't see how. `CMAKE_CURRENT_SOURCE_DIR` is the full path.

jmirabel:
> I don't see how. `CMAKE_CURRENT_SOURCE_DIR` is the full path.

Two paths which differs only from a character that matches your regex will generate the same target. But, you are right, in this case, there is no parent/child relation between the two.

PR #429, opened 2020-08-10
Fix CMake warning in component.

jmirabel:


jmirabel:
This is related to https://github.com/jrl-umi3218/jrl-cmakemodules/pull/417

PR #430, opened 2020-08-10
Boost: don't export python, fix #422, fix #423

### LINKED ISSUE #422: Boost::Python export on 16.04

### LINKED ISSUE #423: Boost::Python export on Arch

nim65s:
Hi,

I checked this workaround on 16.04 & arch :  bulding eigenpy then hpp-fcl worked fine.


jcarpent:
I think we should only export for some Boost components when doing so. No all of them have to be exported.

jmirabel:
I think we shouldn't use this macro for anything else than Boost Python.

For other component, use
- `ADD_PROJECT_DEPENDENCY(Boost COMPONENTS components REQUIRED)` if you want some components to be exported and
- `find_package(Boost COMPONENTS components REQUIRED)` otherwise.

jmirabel:
Maybe this macro should become deprecated in favor of `SEARCH_FOR_BOOST_PYTHON`.

jcarpent:
I do agree that we should introduce `SEARCH_FOR_BOOST_PYTHON` and everything will be fine.
But we also need to export Boost.Python, for instance in Pinocchio, the Python parser relies on it.

jmirabel:
>  But we also need to export Boost.Python, for instance in Pinocchio, the Python parser relies on it.

In general, you do not need to export it. This is a specific case. I suggest either to do it in Pinocchio or, if done in the repo, to do it only when explicitly asked.

nim65s:
the main eigenpy lib also needs it: https://github.com/stack-of-tasks/eigenpy/blob/master/src/eigenpy.cpp#L35

nim65s:
So we agreed on 
1. introducing `SEARCH_FOR_BOOST_PYTHON`
2. deprecating `SEARCH_FOR_BOOST`

Right ?
I can handle this if necessary. But shouldn't we merge this PR until then ?

jcarpent:
We do agree. Please go ahead on top of this PR I would say.

nim65s:
This doesn't work on python 2 / 18.04. I'm trying to avoid committing space changes, which made me miss an indentation issue, which broke my view of the logic, and the logic is wrong here. I'll directly work on `SEARCH_FOR_BOOST_PYTHON` then.

PR #431, opened 2020-08-14
Fix Windows issue

jcarpent:


python.cmake
```diff
@@ -400,6 +400,7 @@ ENDMACRO()
 MACRO(PYTHON_BUILD MODULE FILE)
   # Regex from IsValidTargetName in CMake/Source/cmGeneratorExpression.cxx
   STRING(REGEX REPLACE "[^A-Za-z0-9_.:+-]" "_" compile_pyc "compile_pyc_${CMAKE_CURRENT_SOURCE_DIR}")
+  STRING(REGEX REPLACE ":" "_" compile_pyc "${compile_pyc}")
```
jmirabel, line 403: The two lines above can be merged into
```
   STRING(REGEX REPLACE "[^A-Za-z0-9_.+-]" "_" compile_pyc "compile_pyc_${CMAKE_CURRENT_SOURCE_DIR}")
```

jcarpent: Thanks for the suggestion. I'm not very familiar in the REGEX, so I will adopt your solution then.

PR #432, opened 2020-08-14
Simplify REGEX

jcarpent:


jcarpent:
Following suggestions made in #431.

PR #437, opened 2020-08-17
search boost python, fix #422, fix #423

nim65s:
Replace #430 

It looks like this gives the right results in those cases:
```
16.04/py2       Boost_PYTHON_LIBRARY = /usr/lib/x86_64-linux-gnu/libboost_python-py27.so
16.04/py3       Boost_PYTHON_LIBRARY = /usr/lib/x86_64-linux-gnu/libboost_python-py35.so
18.04/py2       Boost_PYTHON_LIBRARY = /usr/lib/x86_64-linux-gnu/libboost_python-py27.so
18.04/py3       Boost_PYTHON_LIBRARY = /usr/lib/x86_64-linux-gnu/libboost_python-py36.so
20.04/py3       Boost_PYTHON_LIBRARY = /usr/lib/x86_64-linux-gnu/libboost_python38.so
archlinux/py2   Boost_PYTHON_LIBRARY = /usr/lib/libboost_python27.so
archlinux/py3   Boost_PYTHON_LIBRARY = /usr/lib/libboost_python38.so
buster/py2      Boost_PYTHON_LIBRARY = /usr/lib/x86_64-linux-gnu/libboost_python27.so
buster/py3      Boost_PYTHON_LIBRARY = /usr/lib/x86_64-linux-gnu/libboost_python37.so
centos7/py2     Boost_PYTHON_LIBRARY = /usr/lib64/libboost_python-mt.so
centos7/py3     Boost_PYTHON_LIBRARY = /usr/lib64/libboost_python3-mt.so
fedora28/py2    Boost_PYTHON_LIBRARY = /usr/lib64/libboost_python.so
fedora28/py3    Boost_PYTHON_LIBRARY = /usr/lib64/libboost_python3.so
fedora31/py2    Boost_PYTHON_LIBRARY = /usr/lib64/libboost_python27.so
fedora31/py3    Boost_PYTHON_LIBRARY = /usr/lib64/libboost_python37.so
stretch/py2     Boost_PYTHON_LIBRARY = /usr/lib/x86_64-linux-gnu/libboost_python-py27.so
stretch/py3     Boost_PYTHON_LIBRARY = /usr/lib/x86_64-linux-gnu/libboost_python-py35.so
```

This is still a draft, as I need to check a few other things before merging.

PR #438, opened 2020-08-20
Multi-python setup

nim65s:
Hi,

This PR is based on #437, and implement a way to install Boost.Python bindings for several interpreter version simultaneously.

This is still a work in progress: I still need to provide a better description and use examples, before we can efficiently discuss on this topic.


jmirabel:
@nim65s I think you should document somewhere that the main motivation for this is compilation on robotpkg.

I wonder whether this should be exposed to the user or not.

nim65s:
The main motivation for this work was to allow an user or a package manager to install a package for several python version simultaneously. This was meant to ease transition between Python 2 and Python 3. 

But actually, the "added complexity / usefulness" ratio is pretty bad. Also, the python 2 -> 3 transition is already complete in many places, so this PR is becoming obsolete.

PR #440, opened 2020-08-24
Search first for Boost in SEARCH_FOR_BOOST_PYTHON

jcarpent:

PR #442, opened 2020-08-25
Add dedicated USE_BOOST_DEFAULT_OPTIONS

jcarpent:
Compared to #440, this PR introduces a dedicated macro USE_BOOST_DEFAULT_OPTIONS with some minor fix. This will help for the transition.


boost.cmake
```diff
@@ -36,6 +36,28 @@ IF(CMAKE_VERSION VERSION_LESS "3.12")
...
+    FIND_PACKAGE(Boost ${BOOST_REQUIRED})
+  ENDIF(NOT Boost_FOUND)
+  STRING(REPLACE "_" "." Boost_SHORT_VERSION ${Boost_LIB_VERSION})
+  SET(Boost_USE_STATIC_LIBS OFF)
+  SET(Boost_USE_MULTITHREADED ON)
+  IF("${Boost_SHORT_VERSION}" VERSION_GREATER "1.70" OR "${Boost_SHORT_VERSION}" VERSION_EQUAL "1.70")
+    SET(BUILD_SHARED_LIBS ON) 
+    SET(Boost_NO_BOOST_CMAKE ON) 
+  ENDIF("${Boost_SHORT_VERSION}" VERSION_GREATER "1.70" OR "${Boost_SHORT_VERSION}" VERSION_EQUAL "1.70")
+ENDMACRO(USE_BOOST_DEFAULT_OPTIONS)
```
jmirabel, line 59: The name of the function is not clear. If the function only setup the default options, then there shouldn't be a `find_package` inside.

Is this an internal function or is it meant to be used outside ?

If it is meant to be used outside, I would prefer to move the find_package outside, if possible. If not, then you should rename the macro.

jcarpent: I removed any dependency to find_package. I did think it would work but it seems to work well.

jmirabel: Thanks.

jmirabel: In the past, macro `search_for_boost` added lines to the configured Config.cmake. Is this still needed ?

jcarpent: Yes, you're right. This even mandatory.

jcarpent: @jmirabel Maybe, we should add it to `ADD_PROJECT_DEPENDENCY` which will detect the special case of Boost. Or do you think we add it if we include boost.cmake anyway?

jcarpent: I've added a new function, which should be called when required.

jmirabel: I prefer to let users explicitly write it in their CMakeLists so a macro, as you did, is better in my opinion.

PR #445, opened 2020-08-25
Revert "Fix settings on Windows"

jcarpent:
Reverts jrl-umi3218/jrl-cmakemodules#444
I was going too fast for this fix. The wrong behavior on Windows is coming from another point.


jmirabel:
I suggest you to use branch *windows* until you are satisfied.

jcarpent:
The thing is that we are merging some new features without really testing it for the whole bunch of OS we are trying to support. It would be relevant at this stage to think about how we can setup such kind of tests.

jmirabel:
I think we should work on separate branches with clear naming until the downstream package is fixed. And then merge into master.

jcarpent:
Yes and no. We should avoid changing the master branch when it implies that we reduce the number of OS covered. For that aim, we should enforce the CI.

jmirabel:
> For that aim, we should enforce the CI.

I don't think this is impossible.

> We should avoid changing the master branch when it implies that we reduce the number of OS covered.

Yes, and this implies not pushing commits to master while we aren't done with a defined bunch of changes. We have to use a staging area.

PR #448, opened 2020-10-01
Use CMAKE_SOURCE_DIR instead of PROJECT_SOURCE_DIR

jmirabel:
because CMAKE_SOURCE_DIR is always defined while PROJECT_SOURCE_DIR
is only defined after `project` has been called.

### DIFF
--- a/boost.cmake
+++ b/boost.cmake
@@ -32,7 +32,7 @@ FUNCTION(SEARCH_FOR_BOOST_COMPONENT boost_python_name found)
 ENDFUNCTION(SEARCH_FOR_BOOST_COMPONENT boost_python_name found)
 
 IF(CMAKE_VERSION VERSION_LESS "3.12")
-  SET(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/boost ${CMAKE_MODULE_PATH})
+  SET(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/boost ${CMAKE_MODULE_PATH})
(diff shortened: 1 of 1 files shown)


jcarpent (approved):

nim65s (approved):

jmirabel:
The only thing I wonder is what happen in a case such as a project including another project :
- `foo/CMakeLists.txt` calls `project(foo)` and `add_subdirectory('bar')`
- `foo/bar/CMakeLists.txt` calls `project(bar)` and uses this module.

I suspect PROJECT_SOURCE_DIR will point to *foo/bar*, while CMAKE_SOURCE_DIR will point to *foo*. Maybe it is better to use *CMAKE_CURRENT_SOURCE_DIR*.

boost.cmake
```diff
@@ -32,7 +32,7 @@ FUNCTION(SEARCH_FOR_BOOST_COMPONENT boost_python_name found)
 ENDFUNCTION(SEARCH_FOR_BOOST_COMPONENT boost_python_name found)
 
 IF(CMAKE_VERSION VERSION_LESS "3.12")
-  SET(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/boost ${CMAKE_MODULE_PATH})
+  SET(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/boost ${CMAKE_MODULE_PATH})
```
gergondet, line 35: `${CMAKE_CURRENT_LIST_DIR}/boost' will work correctly 100% of the time

In fact, in general it could be nice to pass through the jrl-cmakemodules to replace all assumptions of the `cmake` folder with `${CMAKE_CURRENT_LIST_DIR}`

PR #449, opened 2020-10-01
Remove assumption of `cmake` directory.

jmirabel:
Follows @gergondet suggestion in https://github.com/jrl-umi3218/jrl-cmakemodules/pull/448#discussion_r498334197


jcarpent (approved):

jmirabel:
> You missed one in doxygen/Doxyfile.in

This file does not exist so I guess I do not understand what you mean.

gergondet:
@jmirabel my bad, I was looking at an old version of the modules

jcarpent:
Just to mention that this PR provides bugs, see #450 and #451. I would suggest testing it before merging.
Maybe, it should be nice to setup a more complete CI and testings.

jmirabel:
>  Maybe, it should be nice to setup a more complete CI and testings.

Any idea on how to do that ? Using the script mode is definitely not sufficient. Maybe a set of dummy elementary projects to be compiled ?

jcarpent:
Yes, for the Inria/LAAS side, I would suggest testing EigenPy, SoT and Pinocchio will already remove a lot of bugs.

jmirabel:
Testing on EigenPy, SoT and Pinocchio is time-consuming because you want to compile to be sure it works. i don't think it is a good idea.

PR #450, opened 2020-10-05
Fix usage of CMAKE_CURRENT_LIST_DIR

jcarpent:


jcarpent:
This fixes a bug introduced in #449.

PR #453, opened 2020-10-06
Fx path for Doxygen

jcarpent:


jcarpent:
Related to failures on https://github.com/humanoid-path-planner/hpp-fcl/pull/197

jcarpent:
Seems working. I will not merge it.

base.cmake
```diff
@@ -161,13 +161,15 @@ SET(CMAKE_PROJECT_${PROJECT_NAME}_INCLUDE "${CMAKE_CURRENT_LIST_DIR}/post-projec
...
-VARIABLE_WATCH(CMAKE_CURRENT_SOURCE_DIR SETUP_PROJECT_FINALIZE_HOOK)
+VARIABLE_WATCH(CMAKE_CURRENT_LIST_DIR SETUP_PROJECT_FINALIZE_HOOK)
 FUNCTION(SETUP_PROJECT_FINALIZE_HOOK VARIABLE ACCESS)
   IF("${${VARIABLE}}" STREQUAL "")
+    SET(CMAKE_CURRENT_LIST_DIR ${PROJECT_JRL_CMAKE_MODULE_DIR})
     SETUP_PROJECT_FINALIZE()
     IF(PROJECT_USE_CMAKE_EXPORT)
       SETUP_PROJECT_PACKAGE_FINALIZE()
     ENDIF()
+    SET(CMAKE_CURRENT_LIST_DIR "") # restore value
```
jmirabel, line 172: Not sure to understand: 
- why the two `SET` are in the loop,
- and the *restore value*, which does not seem to *restore* anything.

jcarpent: SETUP_PROJECT_FINALIZE_HOOK is called when CMAKE_CURRENT_LIST_DIR is changed, i.e. when it goes to "".
I use the restored value of CMAKE_CURRENT_LIST_DIR (${PROJECT_JRL_CMAKE_MODULE_DIR}), in order to let SETUP_PROJECT_FINALIZE and SETUP_PROJECT_PACKAGE_FINALIZE to get access to the right value, and not the empty string.

jmirabel: Thanks for the explanation. May it is worth adding this in a comment as it is not straightforward.

PR #455, opened 2020-10-07
allow major.minor versions

nim65s:
Hi,

Some packages are using only 2 numbers to describe the version. For example, in https://github.com/loco-3d/multicontact-locomotion-planning/, the configuration step is failing with:

```
-- Package version (git describe): 1.2-603-g0fb4
CMake Error at CMakeLists.txt:31 (PROJECT):
  VERSION "1.2-603-g0fb4." format invalid.
```

If we agree to support packages with `vMajor.minor` or `vMajor.minor.patch`, we should fix this behavior. This is what this PR is doing. Otherwise, I think we should document that version numbers must have 3 components.

PS: the versions used as examples in the doc only have 2 components.

cc @stonneau


jmirabel:
> If we agree to support packages with `vMajor.minor` or `vMajor.minor.patch`

I agree.

jcarpent (approved):

PR #459, opened 2020-10-29
Move pkg-config.cmake include to post-project

gergondet:
The motivation for this is that the file includes shared-library.cmake which then `include(CMakeFindBinUtils)` and this messes up internal CMake variables so that it breaks toolchains for cross-compilation scenarios.

I think this only breaks anything if one does `add_required|optional_dependency` before calling `project(...)` and I don't think anyone does this.


nim65s (approved):

jcarpent (approved):
Fine for me too.

PR #461, opened 2020-11-07
Fix issue in SEARCH_FOR_BOOST_PYTHON

jcarpent:


boost.cmake
```diff
@@ -113,6 +100,7 @@ MACRO(SEARCH_FOR_BOOST_PYTHON)
   # If boost-python has not been found, warn the user, and look for just "python"
   IF(NOT BOOST_PYTHON_FOUND)
     MESSAGE(WARNING "Impossible to check Boost.Python version. Trying with 'python'.")
+    SET(BOOST_PYTHON_NAME "python")
```
nim65s, line 103: I don't understand. The one on line 86 / 73 is not enough ?

jcarpent: Just check the message you wrote above: `Trying with python`

jcarpent: You are never testing with `python`.

nim65s: you are adding a `SET(BOOST_PYTHON_NAME "python")`, but there is already a `SET(BOOST_PYTHON_NAME "python")` in the second line of the macro to initialize this variable.
So yes, I think that when nothing change `$BOOST_PYTHON_NAME`, we keep the default from that line, which is `python`, so that comment looks good to me

jcarpent: Fine. I did not see it.

jcarpent:
```suggestion
```

PR #462, opened 2020-11-12
[boost] Takes into account Boost version to choose the right link.

### LINKED ISSUE #413: Boost 1.72

olivier-stasse:
### Problem
Fix for the Boost::Python link in case of APPLE.
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/002cd0ff517ec4a31fabe6ec7fba759ada99ea51/boost.cmake#L260
replaces ${Boost_PYTHON_LIBRARIES} by Boost::Python leading to a linking issue.

### Context
cmake 3.18.4
boost 1.73
macOS Catalina 10.15.7

### Approach of the patch
Test Boost_Short_VERSION if below 1.70 then uses previous approach otherwise uses ${Boost_PYTHON_LIBRARY}. 
Supersede #414 in trying to fix #413


jcarpent (commented):
Hi @olivier-stasse,
It is a bit weird because on my side, I did not encounter such troubles. How did you install boost?

olivier-stasse:
Boost was installed with`brew install boost` and it was pulled through a bottle.
Just redoing the test, boost version 1.74 was installed and I tried to recompile `dynamic-graph-python`which is using jrl-cmakemodules topic/dynamic_graph_cpp11 commit 9dcde03a I have the following error:
`
CMake Error at src/dynamic_graph/CMakeLists.txt:5 (ADD_LIBRARY):
  Target "wrap" links to target "-Wl,-undefined,dynamic_lookup,Boost::python"
  but the target was not found.  Perhaps a find_package() call is missing for
  an IMPORTED target, or an ALIAS target is missing?
`
I tried with colcon:
`
colcon build --merge-install --packages-select dynamic-graph-python  --cmake-args '  -DCMAKE_CXX_STANDARD=14' ' -DPYTHON_EXECUTABLE=/usr/local/opt/python@3.8/bin/python3.8'
`
I have the same error  with a direct call to cmake:
`
cmake ../../src/sot/dynamic-graph-python -DCMAKE_CXX_STANDARD=14
`

olivier-stasse:
I did recompile eigenpy which is using the macro and is not failing.
So I will investigate what are the differences.
Thanks.

olivier-stasse:
Ok, I understood the issue. This PR is useless.

PR #464, opened 2020-11-13
[Header] Fix wrong use of CMAKE_CURRENT_LIST_DIR

nim65s:
One more fix related to #449

I don't understand why we missed this one, but it was clearly breaking CI on  dynamic-graph: https://gitlab.laas.fr/stack-of-tasks/dynamic-graph/-/jobs/99322


jcarpent (approved):
Yes, because nobody is using this file

jcarpent:
By the way:
```cmake
  # This header, unlike the previous one is *not* installed and is generated
  # in the top-level directory of the build tree.
  # Therefore it must not be included by any distributed header.
```

nim65s:
Got it, thanks !

PR #471, opened 2021-03-12
Topic/msvc local debugger

aescande:
The primary goal of this PR is to extend what's possible to tune for the local debugger options in Visual Studio through the use of the macro `GENERATE_MSVC_DOT_USER_FILE`. In particular, it is now possible to specify the command and working directory associated with a project.
 - Previous use: `GENERATE_MSVC_DOT_USER_FILE(<name> [<additional_path>])`  (still valid)
 - New possible syntax: `GENERATE_MSVC_DOT_USER_FILE(NAME <name> [COMMAND <command>]  [COMMAND_ARGS <args>]  [WORKING_DIRECTORY <dir>] [ADDITIONAL_PATH <additional_path>])`

In passing the macro `GET_MSVC_VERSION` (which I think is only being used by  `GENERATE_MSVC_DOT_USER_FILE`) is deprecated, for two reasons:
 - It relies on the variables `MSVCXX` whose usage is discouraged by [CMake's documentation](https://cmake.org/cmake/help/v3.8/variable/MSVC14.html).
 - It overwrites the `MSCV_VERSION` variable supplied by CMake.

A `GET_MSVC_TOOLS_VERSION` is introduced for replacement.


msvc-specific.cmake
```diff
@@ -29,6 +31,7 @@
 #  - or newer than MSVC 14.0
 #
 MACRO(GET_MSVC_VERSION)
+  message("This macro is deprecated. Use GET_MSVC_TOOLS_VERSION instead." DEPRECATION)
```
jmirabel, line 34: Just a minor comment. I have always seen `message` used like `message(mode message)` and the documentation states the API of message is like this.

If you called it this way, I am sure it works. But it has puzzled me enough for me to check the doc. Do you have a specific preference for this one ?

aescande: No preference. Previous `message` in the files were like this, so I kept it coherent.
I switched the order in 5eef6a3

jmirabel:
Thank you @aescande. I am not experienced enough with MSVC to do a more in-depth review so it is probably better if someone else merge this.

gergondet (approved):

Issue #470, opened 2021-03-12
Doxygen error with TAGFILES


olivier-stasse:
Symptoms:
=========
In `sot-core` building the documentation with `make doc` generates the following error:

`warning: Invalid value for 'TAGFILES' tag at line 25, file Doxyfile: Values in list '"/usr/local/share/doc/example-robot-data/doxygen-html/example-robot-data.doxytag"="/usr/local/share/doc/example-robot-data/doxygen-html" "/Volumes/Devel/sot_ws/install/share/doc/dynamic-graph-python/doxygen-html/dynamic-graph-python.doxytag"="/Volumes/Devel/sot_ws/install/share/doc/dynamic-graph-python/doxygen-html"' are not properly space or comma separated
error: Illegal format for option TAGFILES, no name specified for item '=/usr/local/share/doc/example-robot-data/doxygen-html'
error: Illegal format for option TAGFILES, no name specified for item '=/Volumes/Devel/sot_ws/install/share/doc/dynamic-graph-python/doxygen-html'
`

Tested with doxygen 1.9.1, MacOS Catalina 10.15.7

Diagnostic:
=========
It is due to the Doxyfile having the following line:
`TAGFILES =  "/usr/local/share/doc/example-robot-data/doxygen-html/example-robot-data.doxytag"="/usr/local/share/doc/example-robot-data/doxygen-html" "/Volumes/Devel/sot_ws/install/share/doc/dynamic-graph-python/doxygen-html/dynamic-graph-python.doxytag"="/Volumes/Devel/sot_ws/install/share/doc/dynamic-graph-python/doxygen-html"`

Removing the quotes fixes the problem.

According to https://www.doxygen.nl/manual/external.html there is no quote in TAGFILES.

It is possible to modify the following line ? 
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/7e797cbcae0947acce802c0bc63765a2b7c75828/doxygen.cmake#L661

jmirabel:
Yes, this line can be modifies. However, dues to the fact that paths can have spaces (and assuming they won't contain the equal sign), I think it is better to go for the syntax mentioned in https://www.doxygen.nl/manual/config.html#cfg_tagfiles
```
[...]

PR #472, opened 2021-03-12
[doxygen] Fix syntax for TAGFILES. 

olivier-stasse:
Remove quotes at the left and right of the = sign.
Fix https://github.com/jrl-umi3218/jrl-cmakemodules/issues/470


jmirabel:
While here, @nim65s the documentations published at https://gepettoweb.laas.fr/doc/ aren't linked to one another. To generate the links from one doc to another, the only thing to do is to update the part after the equal sign contains the URL to the doxytag online.

This probably requires a slight modification of this portion of the code so that you can point to the correct doxytag. I am sure you know how to achieve this.

nim65s:
Thanks for the pointer, @jmirabel, I'll handle this :)
And thanks for the fix @olivier-stasse !

PR #473, opened 2021-04-06
Add simple way to retrieve OpenMP

jcarpent:
It looks like the default script is not working correctly on MAC.


jcarpent:
I will merge this new FindOpenMP as it is not blocking.
I may consider updating it in the future.

Issue #478, opened 2021-05-21
CHECK_MINIMAL_CXX_STANDARD raises warning since macro is not available


wxmerkt:
I am using projects builds with `jrl-cmakemodules` in downstream projects that do not use the `jrl-cmakemodules`. Hence, the lines added to check minimal C++ standards in the `projectConfig.cmake` do not check the version and instead raise a warning (the macro isn't included/available):

```cmake
if(COMMAND CHECK_MINIMAL_CXX_STANDARD)
  CHECK_MINIMAL_CXX_STANDARD(14)
else()
  MESSAGE(WARNING "This dependency requires C++ >= 14")
endif()
```

Is there a recommended solution? E.g. to include the macro in the generated `projectConfig`?

wxmerkt:
Are other macros used in the generated `Config`? If not, we can just fix the exported part in `package-config.cmake`

nim65s:
This macro requires a c++ file to work.

We also need to ship stuff in find-external when they are used, ref. eg. https://github.com/loco-3d/ndcurves/pull/50

wxmerkt:
Thanks for the insight - I've been working on a potential solution locally and will continue Monday and keep you posted

wxmerkt:
Resolved via #479

PR #479, opened 2021-05-27
Fix CHECK_MINIMAL_CXX_STANDARD in exported CMake configs

wxmerkt:
This PR provides a fix for #478 in two parts:

1. It handles a case in `cxx-standard.cmake` that was not yet included: In case a dependent project or a workspace sets CMAKE_CXX_STANDARD before `find_package` of an exported project it may also require checking the standard
2. It includes an adapted full logic to check configurations and adapt, if enforced, in the generated `projectConfig.cmake`. This resolves the initial issue described in #478 when the consumer project does not have a full copy of this jrl-cmakemodules loaded.

Would love to hear thoughts on this!


jcarpent (approved):

cxx-standard.cmake
```diff
@@ -64,7 +64,10 @@ macro(CHECK_MINIMAL_CXX_STANDARD STANDARD)
     if(_COMPILER_DEFAULT_CXX_STANDARD EQUAL 199711
         OR (_COMPILER_DEFAULT_CXX_STANDARD EQUAL 201103 AND _MINIMAL_CXX_STANDARD GREATER 11)
         OR (_COMPILER_DEFAULT_CXX_STANDARD EQUAL 201402 AND _MINIMAL_CXX_STANDARD GREATER 14)
-        OR (_COMPILER_DEFAULT_CXX_STANDARD EQUAL 201703 AND _MINIMAL_CXX_STANDARD GREATER 17))
+        OR (_COMPILER_DEFAULT_CXX_STANDARD EQUAL 201703 AND _MINIMAL_CXX_STANDARD GREATER 17)
```
gergondet, line 67: While we are touching this file we could update it for C++20 since the standard has been released

cxx-standard.cmake
```diff
@@ -64,7 +64,10 @@ macro(CHECK_MINIMAL_CXX_STANDARD STANDARD)
...
+        OR (DEFINED CMAKE_CXX_STANDARD AND CMAKE_CXX_STANDARD LESS 98 AND _MINIMAL_CXX_STANDARD GREATER CMAKE_CXX_STANDARD)
+        OR (DEFINED CMAKE_CXX_STANDARD AND CMAKE_CXX_STANDARD EQUAL 98 AND _MINIMAL_CXX_STANDARD LESS CMAKE_CXX_STANDARD)
```
gergondet, line 69: If we are doing this we could also check for `CMAKE_CXX_STANDARD_REQUIRED` (i.e. the consumer C++ standard cannot be bumped) and error if there's an incompatibility here

package-config.cmake
```diff
@@ -137,11 +137,49 @@ string(REPLACE ";" "\n  " PACKAGE_DEPENDENCIES_FIND_DEPENDENCY "${_PACKAGE_CONFI
...
+        message(AUTHOR_WARNING \"CMAKE_CXX_STANDARD automatically upgraded to $\{CMAKE_CXX_STANDARD\} (from $\{ORIGINAL_CMAKE_CXX_STANDARD\})\")
+      else()
+        message(FATAL_ERROR \"CMAKE_CXX_STANDARD upgrade to >= $\{_MINIMAL_CXX_STANDARD\} required\")
+      endif()
+    endif()
+  else()
+    message(STATUS \"C++ standard sufficient: Minimal required $\{_MINIMAL_CXX_STANDARD\}, currently defined: $\{CMAKE_CXX_STANDARD\}\")
+  endif()
+endif()
+")
```
gergondet, line 182: This looks like a copy paste of the `CHECK_MINIMAL_CXX_STANDARD` macro, I suggest we avoid this by:
- putting the body of the macro in a separate file (i.e. everything after `cmake_parse_arguments(...)` and before `endmacro()` that is included in cxx-standard.cmake
- read the content of this file and append it to the `PACKAGE_EXTRA_MACROS` if needed

Note: after looking more carefully, the bit in the macro that sets `_MINIMAL_CXX_STANDARD` should be out of the included file too

See [here](https://github.com/gergondet/include-in-cmake-macro-example) for an example of what I mean

gergondet (changes requested):
Looks great, thanks!

If we could avoid code duplication that would be even better (and easier to update in the future)

nim65s:
I like the idea of including the C++ code in a string :)

But yes, I agree that we shouldn't duplicate this.

I still think we need a generic solution to ship any file from this module, so if we add this in the future, would we have to revert this ?

Also, if your short-term goal is to avoid a warning, we can add an option to remove this warning.

package-config.cmake
wxmerkt: I agree this should be cleaner. The logic is not a full copy paste and has been adapted (the original one copy-paste did not work). The two key differences are (as far as I can tell/remember since time has passed):

1. The block around `# Check if we need to upgrade the current minimum` has been removed
2. The logic inside the block of `# Check if a non-trivial minimum has been requested` has been adapted (some of the variables won't be available in the packaged version - some we'd like to remain variables and others we'd like to replaced - cf. use of backslashes)

I'll spend some time today to make this cleaner - however due to time constraints and requirements to check the downstream created files with different configurations this may not be completed.

cxx-standard.cmake
wxmerkt: Done - please verify this is as intended

cxx-standard.cmake
wxmerkt: Done

package-config.cmake
wxmerkt: I found a workable solution and tested with several different configurations - thank you for the pointers and feedback. In the process I've also augmented the existing status/warning message output.

wxmerkt:
@gergondet @jcarpent @nim65s Thank you all for your feedback on this PR. I've added more commits to remove the code duplication and capture more edge cases while augmenting the status output. I've tested it locally with different configurations and downstream requirement settings and it worked well. I believe this should be a proper solution to #478 now.

nim65s (approved):
This looks good to me, thanks !

gergondet (approved):

gergondet:
Thanks for the extra work @wxmerkt

PR #480, opened 2021-05-31
Use CMAKE_INSTALL_LIBDIR instead of lib for package config

gergondet:
This PR uses the `CMAKE_INSTALL_LIBDIR` variable to install the CMake configuration files.

This should mostly affect packaging (ping @nim65s )


nim65s:
Thanks for the fix and the headsup :)

This won't cause any issue on my side, as I already have the necessary `-DCMAKE_INSTALL_LIBDIR=lib` everywhere ;)

PR #482, opened 2021-06-08
[test] Add custom target run_tests to run ctest

wxmerkt:
This improves the integration with build tooling and CI systems including colcon and catkin_make_isolated. Without this target, the unit tests defined by the test macro cannot automatically be run with colcon/catkin_make_isolated tooling. Having this target enables us to activate full ROS-CI for CMake projects based on jrl-cmakemodules.(*)

(*) With a caveat of registration of projects from underlay workspaces that happens with some build tools such as colcon but not others such as catkin-tools that is yet to be resolved.

### DIFF
--- a/test.cmake
+++ b/test.cmake
@@ -46,7 +46,7 @@ IF(NOT TARGET build_tests)
 ENDIF()
 
 # Add new target 'run_tests' to improve integration with build tooling
-IF(NOT TARGET run_tests)
+IF(NOT CMAKE_GENERATOR MATCHES "Visual Studio" AND NOT TARGET run_tests)
   ADD_CUSTOM_TARGET(run_tests
     COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -V
     VERBATIM
--- a/test.cmake
+++ b/test.cmake
@@ -45,6 +45,14 @@ IF(NOT TARGET build_tests)
   ADD_CUSTOM_TARGET(build_tests)
 ENDIF()
 
+# Add new target 'run_tests' to improve integration with build tooling
+IF(NOT TARGET run_tests)
+  ADD_CUSTOM_TARGET(run_tests
+    COMMAND ctest
+    VERBATIM
+  )
+ENDIF()
+
 IF(NOT DEFINED ctest_build_tests_exists)
   SET_PROPERTY(GLOBAL PROPERTY ctest_build_tests_exists OFF)
 ENDIF(NOT DEFINED ctest_build_tests_exists)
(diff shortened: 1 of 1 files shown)


wxmerkt:
> But I would really like to understand the design decisions in those tools that lead to the creation of another target for something that was already provided by CMake…

I've done some digging and lots of experimenting but I have not quite been able to trace back all of them and this does not fix all cases but the majority  (notably catkin_tools won't run tests since it does not run the target on non-catkin packages, whereas both catkin_make and colcon will). In either case, a target named `run_tests` was required. Using a custom target depending on the test target only works with colcon not with catkin_make_isolated. Thus, the proposed workaround is the most widely applicable.

I'll use the CMAKE_TEST_COMMAND - thanks for the suggestion. If there are better solutions, please let me know!

gergondet:
Hi @wxmerkt 

Could you check if this works with Visual Studio projects generated by CMake?

In the generated solution, CMake creates a target named `RUN_TESTS` which can be "built" to run the project's unit tests (much like this patch although I'm not sure this runs the exact same command).

I'm wondering how it interacts with this new target.

wxmerkt:
> Could you check if this works with Visual Studio projects generated by CMake?

I don't have Visual Studio on Windows / these projects set up there. Would a solution be to wrap the logic for now with `if(UNIX)` or if not visual studio? Alternatively, I can also append `if(NOT TARGET RUN_TESTS)` to explicitly avoid this case

wxmerkt:
@gergondet Which of the above solutions [(a) wrap with `if(UNIX)` or (b) wrap with `if(NOT TARGET RUN_TESTS)`] would you prefer to be added, and is there anything else blocking a merge?

gergondet:
Hi @wxmerkt 

Sorry for the delay, I had to get on a Windows box to test this out.

Basically:
- CMake does not complain and generates a project with both the `run_tests` and the `RUN_TESTS` "targets"
- Visual Studio will complain that one of the two targets already exists and not load it
- `if(NOT TARGET RUN_TESTS)` does not do anything

I think we can use `if (CMAKE_GENERATOR MATCHES "Visual Studio")` [ref](https://stackoverflow.com/questions/26836361/check-if-generating-a-visual-studio-solution-or-makefile-from-cmake) as `if(UNIX)` might be too restrictive.

wxmerkt:
Thank you for the feedback @gergondet and please apologise for the delay in amending the PR. I just added the extra if-statement to catch Visual Studio. Is the PR now good to go?

jcarpent (approved):

gergondet (approved):

Issue #483, opened 2021-07-09
Automatically incrementing package.xml version if package.xml present


wxmerkt:
With Justin we have been discussing to automate incrementing/adapting the version in `package.xml` when `make release VERSION=x.y.z` is called.

The current idea is to use `_COMPUTE_VERSION_FROM_ROS_PACKAGE_XML_FILE` from `version.cmake` and if the version does not match the version given with `VERSION=` to use `sed` to amend the `package.xml` and then commit it as part of the release process. 

However, before implementing this - would love to hear some feedback if there are other thoughts / suggestions.

jcarpent:
Solved via #484 and #485

PR #484, opened 2021-07-09
[release] Update version in package.xml when running make release

wxmerkt:
Potential solution to #483

### DIFF
--- a/release.cmake
+++ b/release.cmake
@@ -70,7 +70,7 @@ MACRO(RELEASE_SETUP)
           (echo "Updating package.xml to $$VERSION"
            && sed -i \"s!<version>[0-9].[0-9].[0-9]</version>!<version>$$VERSION</version>!g\" package.xml
            && ${GIT} add package.xml
-           && ${GIT} commit -m "Update package.xml version to $$VERSION"
+           && ${GIT} commit -m "release: Update package.xml version to $$VERSION"
            && echo "Updated package.xml and committed") ;
          fi
       && ${GIT} tag -s v$$VERSION -m "Release of version $$VERSION."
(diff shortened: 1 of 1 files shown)


release.cmake
```diff
@@ -65,6 +65,14 @@ MACRO(RELEASE_SETUP)
       && ! test x$$VERSION = x
         || (echo "Please set a version for this release" && false)
       && cd ${PROJECT_SOURCE_DIR}
+      # Update version in package.xml if it exists
+      && if [ -f "package.xml" ]; then
+          (echo "Updating package.xml to $$VERSION"
+           && sed -i \"s!<version>[0-9].[0-9].[0-9]</version>!<version>$$VERSION</version>!g\" package.xml
+           && ${GIT} add package.xml
+           && ${GIT} commit -m "Update package.xml version to $$VERSION"
```
jcarpent, line 73:
```suggestion
           && ${GIT} commit -m "release: Update package.xml version to $$VERSION"
```

jcarpent (approved):
Thanks a lot @wxmerkt. This looks nice to me.
I made a simple suggestion ;)

nim65s (approved):
I guess the assumption that the `version` tag is on a single line is good enough :)

release.cmake
```diff
@@ -65,6 +65,14 @@ MACRO(RELEASE_SETUP)
       && ! test x$$VERSION = x
         || (echo "Please set a version for this release" && false)
       && cd ${PROJECT_SOURCE_DIR}
+      # Update version in package.xml if it exists
+      && if [ -f "package.xml" ]; then
+          (echo "Updating package.xml to $$VERSION"
+           && sed -i \"s!<version>[0-9].[0-9].[0-9]</version>!<version>$$VERSION</version>!g\" package.xml
```
jmirabel, line 71: Don't you need to add some `+` to the regex to support version `10.234.15423542` ?

wxmerkt: Great catch, thank you. I just added a commit to fix it

jmirabel (approved):

PR #485, opened 2021-07-30
Handle BSD sed when updating package.xml

jcarpent:
Sed on OSX and BSD systems is different from Linux version.
I've adapted the sed regular expression to handle these systems.
@nim65s @jmirabel @wxmerkt Could you check on your side I did not break the behaviour on Linux systems?

### DIFF
--- a/release.cmake
+++ b/release.cmake
@@ -69,7 +69,7 @@ MACRO(RELEASE_SETUP)
       # Update version in package.xml if it exists
       && if [ -f "package.xml" ]; then
           (echo "Updating package.xml to $$VERSION"
-           && sed -i ${SED_I_ARG} -E \"s|<version>.*</version>|<version>$$VERSION</version>|g\" package.xml
+           && sed -i ${SED_I_ARG} \"s|<version>.*</version>|<version>$$VERSION</version>|g\" package.xml
            && ${GIT} add package.xml
            && ${GIT} commit -m "release: Update package.xml version to $$VERSION"
            && echo "Updated package.xml and committed") ;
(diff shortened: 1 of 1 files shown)


wxmerkt:
I just tried it on Ubuntu and it fails:
```
sed: can't read s|<version>[0-9].[0-9].[0-9]</version>|<version>1.2.3</version>|g: No such file or directory
```

Do we have to add the "+" again as well as @jmirabel pointed out in the other PR?

jcarpent:
In fact, it seems the sed command is failing on Linux systems.
I think the best way would be to split between systems.

jmirabel:
Sereval potential solutions:
- @nim65s is `awk` more portable ?
- I don't think we have to care for what is inside the version tag so no need for extended regex. On my system, `sed -i 's/<version>.*<\/version>/foo/'` suffice.

jcarpent:
I think we are done now. I've added the comment of @jmirabel.
Thanks @wxmerkt and @jmirabel for your help.

release.cmake
```diff
@@ -68,7 +69,7 @@ MACRO(RELEASE_SETUP)
       # Update version in package.xml if it exists
       && if [ -f "package.xml" ]; then
           (echo "Updating package.xml to $$VERSION"
-           && sed -i --expression=\"s!<version>[0-9]\\+.[0-9]\\+.[0-9]\\+</version>!<version>$$VERSION</version>!g\" package.xml
+           && sed -i ${SED_I_ARG} -E \"s|<version>.*</version>|<version>$$VERSION</version>|g\" package.xml
```
jmirabel, line 72: I think `-E` is useless. If you want to have the same for both mac and linux:
```suggestion
           && sed -i.bak \"s|<version>.*</version>|<version>$$VERSION</version>|g\" package.xml
           && rm package.xml.bak
```

jcarpent: You're correct. I apply the change.

PR #486, opened 2021-07-30
Extract NUMPY_VERSION

jcarpent:


jcarpent:
I will merge as it should not impact someone.

PR #501, opened 2021-11-24
[sdformat] check for different sdformat packages, and add project dependency when found

proyan:


jcarpent (changes requested):
Could you automatize the process, meaning it would be easy to handle just a list of versions?
In addition, I would suggest relying on the latest version first.

proyan:
The last commit should handle both your suggestions.

jcarpent (approved):
Much more elegant ;) Thanks a lot.

nim65s:
Maybe we could add a `REQUIRED` argument, that raise an error on reaching the end of the loop ?

proyan:
@nim65s That is a good suggestion. https://github.com/jrl-umi3218/jrl-cmakemodules/pull/501/commits/d9df3999735feb0402d88c142fd517a7ecb7d99f should take care of that

nim65s (approved):
thanks !

PR #504, opened 2022-02-16
config: Avoid type-limits warning by performing signed comparison

wxmerkt:

PR #505, opened 2022-02-16
Revert "config: Avoid type-limits warning by performing signed comparison"

wxmerkt:
Reverts jrl-umi3218/jrl-cmakemodules#504

Turns out I had only tested this with its use in C++ code, but not in macros (`#if HPP_FCL_VERSION_AT_LEAST(1,1,1)`). We can't cast as a macro, hence the revert.

### DIFF
--- a/config.hh.cmake
+++ b/config.hh.cmake
@@ -13,15 +13,13 @@
 # define @LIBRARY_NAME@_MINOR_VERSION @PROJECT_VERSION_MINOR_CONFIG@
 # define @LIBRARY_NAME@_PATCH_VERSION @PROJECT_VERSION_PATCH_CONFIG@
 
-// Since i <= 0 is always true, we'll get a compiler warning if major, minor or patch versions are equal to 0.
-// To avoid throwing an error on otherwise valid code, we explicitly perform a signed <= comparison
-#define @LIBRARY_NAME@_VERSION_AT_LEAST(major, minor, patch) (@LIBRARY_NAME@_MAJOR_VERSION>major || ((int)@LIBRARY_NAME@_MAJOR_VERSION>=(int)major && \
-                                                             (@LIBRARY_NAME@_MINOR_VERSION>minor || ((int)@LIBRARY_NAME@_MINOR_VERSION>=(int)minor && \
-                                                                                                     (int)@LIBRARY_NAME@_PATCH_VERSION>=(int)patch))))
+#define @LIBRARY_NAME@_VERSION_AT_LEAST(major, minor, patch) (@LIBRARY_NAME@_MAJOR_VERSION>major || (@LIBRARY_NAME@_MAJOR_VERSION>=major && \
+                                                             (@LIBRARY_NAME@_MINOR_VERSION>minor || (@LIBRARY_NAME@_MINOR_VERSION>=minor && \
+                                                                                                     @LIBRARY_NAME@_PATCH_VERSION>=patch))))
 
-#define @LIBRARY_NAME@_VERSION_AT_MOST(major, minor, patch) (@LIBRARY_NAME@_MAJOR_VERSION<major || ((int)@LIBRARY_NAME@_MAJOR_VERSION<=(int)major && \
-                                                            (@LIBRARY_NAME@_MINOR_VERSION<minor || ((int)@LIBRARY_NAME@_MINOR_VERSION<=(int)minor && \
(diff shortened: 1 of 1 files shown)


config.hh.cmake
```diff
@@ -13,15 +13,13 @@
...
 # define @LIBRARY_NAME@_PATCH_VERSION @PROJECT_VERSION_PATCH_CONFIG@
 
-// Since i <= 0 is always true, we'll get a compiler warning if major, minor or patch versions are equal to 0.
-// To avoid throwing an error on otherwise valid code, we explicitly perform a signed <= comparison
-#define @LIBRARY_NAME@_VERSION_AT_LEAST(major, minor, patch) (@LIBRARY_NAME@_MAJOR_VERSION>major || ((int)@LIBRARY_NAME@_MAJOR_VERSION>=(int)major && \
-                                                             (@LIBRARY_NAME@_MINOR_VERSION>minor || ((int)@LIBRARY_NAME@_MINOR_VERSION>=(int)minor && \
-                                                                                                     (int)@LIBRARY_NAME@_PATCH_VERSION>=(int)patch))))
+#define @LIBRARY_NAME@_VERSION_AT_LEAST(major, minor, patch) (@LIBRARY_NAME@_MAJOR_VERSION>major || (@LIBRARY_NAME@_MAJOR_VERSION>=major && \
+                                                             (@LIBRARY_NAME@_MINOR_VERSION>minor || (@LIBRARY_NAME@_MINOR_VERSION>=minor && \
+                                                                                                     @LIBRARY_NAME@_PATCH_VERSION>=patch))))
```
jmirabel, line 18: Not sure why this isn't
```

#define @LIBRARY_NAME@_VERSION_AT_LEAST(major, minor, patch) (@LIBRARY_NAME@_MAJOR_VERSION>major || (@LIBRARY_NAME@_MAJOR_VERSION==major && \
                                                             (@LIBRARY_NAME@_MINOR_VERSION>minor || (@LIBRARY_NAME@_MINOR_VERSION==minor && \
                                                                                                     @LIBRARY_NAME@_PATCH_VERSION>=patch))))
```
which will trigger one warning instead of 3.

For the *patch* check, I think the following equivalent code will not trigger a warning
`@LIBRARY_NAME@_PATCH_VERSION==patch || @LIBRARY_NAME@_PATCH_VERSION>patch`

config.hh.cmake
```diff
@@ -13,15 +13,13 @@
...
 
-#define @LIBRARY_NAME@_VERSION_AT_MOST(major, minor, patch) (@LIBRARY_NAME@_MAJOR_VERSION<major || ((int)@LIBRARY_NAME@_MAJOR_VERSION<=(int)major && \
-                                                            (@LIBRARY_NAME@_MINOR_VERSION<minor || ((int)@LIBRARY_NAME@_MINOR_VERSION<=(int)minor && \
-                                                                                                    (int)@LIBRARY_NAME@_PATCH_VERSION<=(int)patch))))
+#define @LIBRARY_NAME@_VERSION_AT_MOST(major, minor, patch) (@LIBRARY_NAME@_MAJOR_VERSION<major || (@LIBRARY_NAME@_MAJOR_VERSION<=major && \
+                                                            (@LIBRARY_NAME@_MINOR_VERSION<minor || (@LIBRARY_NAME@_MINOR_VERSION<=minor && \
+                                                                                                     @LIBRARY_NAME@_PATCH_VERSION<=patch))))
```
jmirabel, line 22: `<=` should never warn, shouldn't it ?

wxmerkt: It warned with clang (I believe 12, maybe 10), I guess it optimised and flipped to >= during compilation (?)

jmirabel: Warning happens at parsing stage. Optimization happens after so can't be flipped.

jmirabel: Ok. It's obvious now. What warns is `<` and not `<=`.

jmirabel: We can add a `@LIBRARY_NAME@_VERSION_EQUALS` macro and make `@LIBRARY_NAME@_VERSION_AT_MOST` rely on `@LIBRARY_NAME@_VERSION_AT_LEAST` and `@LIBRARY_NAME@_VERSION_EQUALS`

wxmerkt: That would make sense - thank you for the suggestions. For now, I've fixed it on the C++ side by converting the passed in values for major, minor, and patch to signed.

PR #506, opened 2022-02-16
Check whether the try_run has succeeded

jcarpent:


cxx-standard.cmake
```diff
@@ -42,7 +42,12 @@ macro(CHECK_MINIMAL_CXX_STANDARD STANDARD)
     try_run(_cxx_standard_run_status _cxx_standard_build_status
       ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR}/cmake/tmp-cxx-standard.cpp
       RUN_OUTPUT_VARIABLE _COMPILER_DEFAULT_CXX_STANDARD)
+    if(_cxx_standard_run_status EQUAL FAILED_TO_RUN OR _cxx_standard_build_status EQUAL FAILED_TO_RUN)
```
nim65s, line 45: I didn't test, but `_cxx_standard_build_status` should be `TRUE` or `FALSE`, not `FAILED_TO_RUN`

jcarpent: You're totally right.

PR #507, opened 2022-03-01
add `ENABLE_COVERAGE` option and `coverage` target

nim65s:
Hi,

This merge coverage data from [lcov](http://ltp.sourceforge.net/coverage/lcov.php) for C++ and [coverage.py](https://github.com/nedbat/coveragepy) for python in a single HTML site in `build/coverage`.

This require both `test` and `install` targets, is I can't find a better way than give installed python files as sources to coverage.py.


coverage.cmake
```diff
@@ -0,0 +1,45 @@
...
+#
+
+#.rst:
+# .. ifmode:: internal
+#
+#   .. variable:: ENABLE_COVERAGE
+#
+#      When this is ON, coverage compiler flags are enabled
+
+option(ENABLE_COVERAGE "enable C++ and python code coverage" OFF)
```
gergondet, line 15:
```suggestion
option(ENABLE_COVERAGE "Enable C++ and Python code coverage" OFF)
```

gergondet, line 15: I think the option shouldn't exist if the compiler doesn't support the `--coverage` flag, the following is probably good enough:

```cmake
if(NOT MSVC)
  option(ENABLE_COVERAGE ...)
else()
  set(ENABLE_COVERAGE OFF)
endif()
```

coverage.cmake
```diff
@@ -0,0 +1,45 @@
...
+    # and then generate HTML by removing the source prefix and python sitelib install prefix
+    add_custom_target(coverage
+      COMMAND lcov --include "${PROJECT_SOURCE_DIR}/\\*" -c -d ${PROJECT_SOURCE_DIR} -o cpp.lcov
+      COMMAND ${PYTHON_EXECUTABLE} -m coverage combine
+      COMMAND ${PYTHON_EXECUTABLE} -m coverage lcov -o python.lcov
+      COMMAND genhtml -p ${PROJECT_SOURCE_DIR} -p "${CMAKE_INSTALL_PREFIX}/${PYTHON_SITELIB}"
+              -o ${_COVERAGE_DIR} cpp.lcov python.lcov
+      BYPRODUCTS .coverage cpp.lcov python.lcov ${_COVERAGE_DIR}
+      WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+      COMMENT "generating code coverage data")
```
gergondet, line 43:
```suggestion
      COMMENT "Generating code coverage data")
```

coverage.cmake
```diff
@@ -0,0 +1,45 @@
...
+
+macro(_SETUP_COVERAGE_FINALIZE)
+  if(ENABLE_COVERAGE)
+    set(_COVERAGE_DIR "${CMAKE_BINARY_DIR}/coverage")
+
+    # use lcov to gether c++ coverage reports from sources,
+    # and coverage.py to gather python coverage reports from installed files,
+    # and then generate HTML by removing the source prefix and python sitelib install prefix
+    add_custom_target(coverage
+      COMMAND lcov --include "${PROJECT_SOURCE_DIR}/\\*" -c -d ${PROJECT_SOURCE_DIR} -o cpp.lcov
```
gergondet, line 36: We should `find_program(LCOV lcov)` to make sure it's installed and errour out if it's not (and `ERROR_COVERAGE` is `ON` of course)

coverage.cmake
```diff
@@ -0,0 +1,45 @@
...
+      COMMAND ${PYTHON_EXECUTABLE} -m coverage combine
+      COMMAND ${PYTHON_EXECUTABLE} -m coverage lcov -o python.lcov
+      COMMAND genhtml -p ${PROJECT_SOURCE_DIR} -p "${CMAKE_INSTALL_PREFIX}/${PYTHON_SITELIB}"
```
gergondet, line 39: Same remark as `lcov` (much less likely to fail if `lcov` is found)

coverage.cmake
```diff
@@ -0,0 +1,45 @@
...
+  if(ENABLE_COVERAGE)
+    set(_COVERAGE_DIR "${CMAKE_BINARY_DIR}/coverage")
+
+    # use lcov to gether c++ coverage reports from sources,
+    # and coverage.py to gather python coverage reports from installed files,
+    # and then generate HTML by removing the source prefix and python sitelib install prefix
+    add_custom_target(coverage
+      COMMAND lcov --include "${PROJECT_SOURCE_DIR}/\\*" -c -d ${PROJECT_SOURCE_DIR} -o cpp.lcov
+      COMMAND ${PYTHON_EXECUTABLE} -m coverage combine
+      COMMAND ${PYTHON_EXECUTABLE} -m coverage lcov -o python.lcov
```
gergondet, line 38: Maybe we don't want/need to run this if there's no Python code or Python tests or the first `lcov` command if there's no C++ coverage, maybe (in `_SETUP_COVERAGE`):
```
set_property(GLOBAL PROPERTY JRL_CMAKEMODULES_HAS_CPP_COVERAGE OFF)
set_property(GLOBAL PROPERTY JRL_CMAKEMODULES_HAS_PYTHON_COVERAGE OFF)
```

Then these properties are set to `ON` if coverage is added to a target (C++ side) or if Python tests are added

test.cmake
```diff
@@ -109,8 +109,15 @@ ENDMACRO(ADD_UNIT_TEST NAME SOURCE)
 #   .. note:: :command:`FINDPYTHON` should have been called first.
 #
 MACRO(ADD_PYTHON_UNIT_TEST NAME SOURCE)
-  ADD_TEST(NAME ${NAME} COMMAND ${PYTHON_EXECUTABLE} "${PROJECT_SOURCE_DIR}/${SOURCE}")
-  SET(PYTHONPATH)
+  IF(ENABLE_COVERAGE)
+    SET(PYTHONPATH "${CMAKE_INSTALL_PREFIX}/${PYTHON_SITELIB}")
+    ADD_TEST(NAME ${NAME}
+      COMMAND ${PYTHON_EXECUTABLE} -m coverage run --branch -p --source=${PYTHONPATH} "${PROJECT_SOURCE_DIR}/${SOURCE}"
+      WORKING_DIRECTORY ${PROJECT_BINARY_DIR})
```
gergondet, line 116: This might be a subtle breaking change when coverage is enabled

gergondet (changes requested):

nim65s:
…

coverage.cmake
nim65s: fixed in #509

coverage.cmake
```diff
@@ -0,0 +1,45 @@
...
+macro(_SETUP_COVERAGE_FINALIZE)
+  if(ENABLE_COVERAGE)
+    set(_COVERAGE_DIR "${CMAKE_BINARY_DIR}/coverage")
+
+    # use lcov to gether c++ coverage reports from sources,
+    # and coverage.py to gather python coverage reports from installed files,
+    # and then generate HTML by removing the source prefix and python sitelib install prefix
+    add_custom_target(coverage
+      COMMAND lcov --include "${PROJECT_SOURCE_DIR}/\\*" -c -d ${PROJECT_SOURCE_DIR} -o cpp.lcov
+      COMMAND ${PYTHON_EXECUTABLE} -m coverage combine
```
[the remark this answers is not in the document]
nim65s: fixed in #509

coverage.cmake
nim65s: fixed in #509, but by checking if `PYTHON_EXECUTABLE` is defined. But your solution looks actually better. I'll implement that.

coverage.cmake
```diff
@@ -0,0 +1,45 @@
...
+#      When this is ON, coverage compiler flags are enabled
+
+option(ENABLE_COVERAGE "enable C++ and python code coverage" OFF)
+
+macro(_SETUP_COVERAGE)
+  if(ENABLE_COVERAGE)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} --coverage")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --coverage")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} --coverage")
+    message(STATUS "Appending code coverage compiler flags: --coverage")
```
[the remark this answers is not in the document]
nim65s: done in #509

test.cmake
nim65s: I already found some of those breaking changes in hpp-fcl, as a python helper script is available in the test source dir, but not in the installed project

gergondet: If you don't mind requiring CMake 3.7 (notably this excludes "stock" CMake in 16.04) you can use [test fixtures](https://cmake.org/cmake/help/latest/prop_test/FIXTURES_REQUIRED.html) to move the generated coverage files after the run. The cleanup test can be something like `${CMAKE_COMMAND} -DPROJECT_BINARY_DIR="${PROJECT_BINARY_DIR}" -DSOURCE=${SOURCE} -P cleanup-python-coverage.cmake` where `cleanup-python-coverage.cmake` is some CMake script that boils down to `mv ${SOURCE}.lcov ${PROJECT_BINARY_DIR}` (but with `${CMAKE_COMMAND} -E`)

If you want to keep backward compatibility you can use a regular dependency between the test and the cleanup test but you will lose the benefits of the fixture (e.g. running the test alone automatically runs the cleanup) but I think it's relatively rare to enable coverage outside of CI.

nim65s: I don't see why we would need to move the generated coverage files

gergondet: Assuming you want to get rid of this `WORKING_DIRERCTORY` then I guess the coverage file would be generated in the default working directory and you'd want to move them to a predictable place so the coverage report can be correctly generated.

But if you are fine with fixing breaking changes introduced by this I don't mind much since I don't use the macro myself.

nim65s: I see. I wasn't considering this `WORKING_DIRECTORY` to be an issue. For me the issue might come from the `--source=${PYTHONPATH}`, because we need to install the project before beeing able to get the coverage. 

The cause is that we need all compiled module to be available, so the source tree is not an option, and we also need the `.py` file, so the build tree is not an option (as, for now, we only have `.pyc` there), and that's why I tried with the install tree.

But in some cases, the install tree might lack python files defined in the source tree and not meant to be installed (eg. https://github.com/humanoid-path-planner/hpp-fcl/blob/devel/test/python_unit/test_case.py)
Maybe this is a design error, as one might expect the test to be available with only the install tree available, eg. after an install from a binary package, so I guess I'll try with this first.

About the changes because of the `WORKING_DIRECTORY`, you're probably right, but my ideas will be more clear once I get an actual error in the wild

PR #509, opened 2022-03-04
add ENABLE_COVERAGE option and coverage target 

nim65s:
I made a wrong push in #507, sorry.


gergondet:
I think (weirdly) we can continue the review on #507 but merge from this PR, would that be ok for you? Otherwise I'll recomment here

nim65s:
Thanks a lot @gergondet for your review ! 
This PR was initially just a very simple prototype to start discussion and make some tests, but not this feel way more polished :D

PR #511, opened 2022-04-07
Fix for weird CMake versions

costashatz:
In some systems (e.g. windows), the cmake version string looks like this `3.17.20032601-MSVC_2`. This causes the check `if(${CMAKE_VERSION} VERSION_GREATER 3.17.6)` to return true, and then cmake fails (because `CheckLinkerFlag` does not exist). This PR does the opposite check `VERSION_LESS` (since `VERSION_GREAT_OR_EQUAL` does not exist). This is affecting my `SpaceVecAlg` and `RBDyn` builds.

### DIFF
--- a/version-script.cmake
+++ b/version-script.cmake
@@ -15,7 +15,9 @@
 
 include(CheckCCompilerFlag)
 # Introduced in 3.18.0 but VERSION_GREAT_OR_EQUAL is not available in CMake 3.1
-if(${CMAKE_VERSION} VERSION_GREATER 3.17.6)
+if(${CMAKE_VERSION} VERSION_LESS 3.18.0)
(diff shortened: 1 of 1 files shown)


gergondet (approved):
From the version name I guess this is the version shipped along with Visual Studio, nice catch!

I will merge now.

costashatz:
Thanks! Can you please update `RBDyn` and `SpaceVecAlg` as well to point to the new master?

Thanks!

gergondet:
> Can you please update RBDyn and SpaceVecAlg as well to point to the new master?

Done and done :)

costashatz:
> Done and done :)

Always top class support! Thanks a ton! Have a nice day!

PR #513, opened 2022-05-03
python: Development might not be required

nim65s:
Hi,

In many places, we need only the Python interpreter (and sometimes the include dirs), but libpythonX.Y.so is not necessary.

Building wheel packages allowing distribution on [PyPI](https://pypi.org) even explicitly forbids it (ref. [PEP513](https://peps.python.org/pep-0513/#libpythonx-y-so-1)). Therefore, it is not available on the [manylinux](https://github.com/pypa/manylinux) distribution.

Building jrl-cmakemodules packages on that distribution is currently not working, because CMake consider the `Development` component not found when `libpythonX.Y.so` is not available.

So I think we need to remove that.

Maybe this is required in some other cases I didn't think about yet, so an opt-out option would be better than just this blunt removal. Any idea about that ?

### DIFF
--- a/python.cmake
+++ b/python.cmake
@@ -48,6 +48,11 @@
 #
 #  Portable suffix of C++ Python modules.
 
+#.rst:
+# .. variable:: PYTHON_COMPONENTS
+#
+#  Required components for python. Default: "Interpreter Development"
+
 IF(CMAKE_VERSION VERSION_LESS "3.2")
     SET(CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/python ${CMAKE_MODULE_PATH})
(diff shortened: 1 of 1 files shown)


jcarpent (changes requested):
I would instead suggest making it an optional option or adding a helper to look for it if needed.

nim65s:
Ok, and would you include or exclude this `Development` component by default ?

jcarpent:
We can exclude it or use a dedicated JRL_CMAKEMODULE_VAR to say that we are on a PyPi manylinux setup, or detect it automatically.
I would go for the latest solution.

nim65s:
Here is a solution that will allow for another PR to introduce a manylinux setup which will directly change the default of this `${PYTHON_COMPONENTS}` variable

jcarpent (approved):

PR #516, opened 2022-07-06
double check PYTHON_INCLUDE_DIR for new cmake versions

nim65s:
Hi,

This PR follows #513, and deals with the case where we need access to python include dirs but not the shared library.

This was already working with CMake < 3.12, so I'm just moving the block dealing with that outside of the CMake version check.


nim65s:
Python3::Development.Module seems to be OK

PR #518, opened 2022-07-14
Modernize if available

nim65s:
Hi,

This provide some ways to activate modern behavior when it is available, while trying not to break old-school stuff.

ref. https://github.com/stack-of-tasks/eigenpy/pull/291


modernize-links.cmake
```diff
@@ -0,0 +1,38 @@
+#
+# Copyright (C) 2022 LAAS-CNRS
+#
+# Author: Guilhem Saurel
+#
+
+#.rst:
+# .. ifmode:: user
+#
+#   .. command:: MODERNIZE_LINK(target SCOPE <INTERFACE|PUBLIC|PRIVATE> TARGETS [targets...] LIBRARIES [libraries...]
```
jcarpent, line 10: I would call it `MODERNIZE_TARGET_LINK_LIBRARIES`. It better shows the link with the CMake original signature.

jcarpent (changes requested):

jcarpent (approved):

PR #523, opened 2022-07-21
PYTHON_EXPORT_DEPENDENCY

nim65s:
Hi,

`FINDPYTHON()` currently only use `find_package()`. This PR deals with the case where we need `ADD_PROJECT_DEPENDENCY()` instead, by adding a `PYTHON_EXPORT_DEPENDENCY` setting.

This also export:
     - `PYTHON_SITELIB`
     - `PYTHON_SOABI`
     - `PYTHON_EXT_SUFFIX`
     - `PYTHON_EXECUTABLE` (not the current string, but the variable dynamically from the `find_dependency()` call)
     - Boost::python if `SEARCH_FOR_BOOST_PYTHON` is called


wxmerkt:
Can we get this merged and updated/released in EigenPy please? We have downstream packages breaking :-(

jcarpent (approved):

PR #527, opened 2022-07-26
python: allow users to pass standard -DPython2_EXECUTABLE / -DPython3_EXECUTABLE

nim65s:
this also allows cases where a `find_package(Python3 REQUIRED)` was successfully run before our `FINDPYTHON()` to avoid misleading messages about Python found then Python not found.


python.cmake
```diff
@@ -104,11 +104,23 @@ MACRO(FINDPYTHON)
...
+          set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
+        ELSEIF(DEFINED Python2_EXECUTABLE)
+          set(PYTHON_EXECUTABLE ${Python2_EXECUTABLE})
+        ELSEIF(DEFINED Python3_EXECUTABLE)
+          set(PYTHON_EXECUTABLE ${Python3_EXECUTABLE})
+        ENDIF()
+      ENDIF()
+
       IF(NOT DEFINED Python_EXCUTABLE)
         SET(Python_EXCUTABLE ${PYTHON_EXECUTABLE})
```
gergondet, line 123: That's not related to the PR but why `EXCUTABLE` instead of `EXECUTABLE` here?

jcarpent: This is a mistake ;). My bad.

nim65s: thanks, let's fix

jcarpent (approved):

jcarpent (approved):

PR #529, opened 2022-07-26
call python macros before find_dependency

nim65s:
Hi again,

this continue work from #525, but call our custom macros for python/numpy before the find_dependency.
https://github.com/stack-of-tasks/eigenpy/runs/7518507336?check_suite_focus=true shows why this is necessary
https://github.com/stack-of-tasks/eigenpy/runs/7520451369?check_suite_focus=true will show if this work :)


jcarpent (approved):

nim65s:
As seen by @jcarpent, actually, this should even allow to remove the find_dependency.
https://github.com/stack-of-tasks/eigenpy/runs/7520546310?check_suite_focus=true will double check that.

PR #535, opened 2022-08-06
Fix export for Python modules

jcarpent:


nim65s:
I think this won't work with CMake < 3.14, where we won't export the `Python::NumPy` target (ref l.101). 
Maybe we could `set(SEARCH_FOR_NUMPY TRUE)` in Config.cmake.in ?

jcarpent:
> I think this won't work with CMake < 3.14, where we won't export the `Python::NumPy` target (ref l.101). Maybe we could `set(SEARCH_FOR_NUMPY TRUE)` in Config.cmake.in ?

Do you want to populate PYTHON_EXPORT_DEPENDENCY_MACROS with this?

PR #538, opened 2022-08-10
Add julia support

fabinsch:
for our package proxqp (https://github.com/jcarpent/proxqp-unofficial) with @jcarpent we are providing a julia example. this example requires to have julia and the pkg `PyCall `installed. here we add the functions to 
- check if the julia binary is found
- check if a certain package (pycall in our case) is installed
- add a julia unittest


test.cmake
```diff
@@ -174,6 +174,19 @@ endmacro(
   NAME
   SOURCE)
 
+# .rst: .. command:: ADD_JULIA_UNIT_TEST (NAME SOURCE [MODULES...])
+#
+# Add a test called `NAME` that runs an equivalent of ``julia ${SOURCE}``.
+#
+macro(ADD_JULIA_UNIT_TEST NAME SOURCE)
+  message(STATUS "${Julia_EXECUTABLE} ${PROJECT_SOURCE_DIR}/${SOURCE}")
```
jcarpent, line 182:
```suggestion
```

julia.cmake
```diff
@@ -0,0 +1,15 @@
...
+#
+macro(JULIA_CHECK_PACKAGE PKG_NAME)
+  set(file_content
+      "try\n@eval import ${PKG_NAME}\nprintln(1)\ncatch\nprintln(0)\nend")
+  file(WRITE ${PROJECT_SOURCE_DIR}/_tmp_cmake_julia/test_julia_package.jl
+       ${file_content})
+  execute_process(
+    COMMAND ${Julia_EXECUTABLE}
+            ${PROJECT_SOURCE_DIR}/_tmp_cmake_julia/test_julia_package.jl
+    OUTPUT_VARIABLE ${PKG_NAME}_found)
```
jcarpent, line 13: Maybe to avoid any clash, we should rename `${PKG_NAME}_found` into `Julia_${PKG_NAME}_found` and add the output variable in the doc.

jcarpent (changes requested):
Thanks @fabinsch for this helpful PR. I made some suggestions.

jcarpent (approved):

PR #540, opened 2022-08-23
Fix configure_package_config_file use

jcarpent:
Related to https://github.com/Simple-Robotics/proxsuite/issues/14


jcarpent:
See https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html

nim65s:
Nice, I didn't knew this, thanks !

PR #544, opened 2022-08-24
Remove useless set(CMAKE_MODULE_PATH

jcarpent:
This set(CMAKE_MODULE_PATH is already included within boost.cmake


jcarpent:
Otherwise, it is using by default the FindBoost.cmake provided by the JRL cmake module.

nim65s (approved):
Ok, then we'll have to do this manually in cases where we'll need it.

jcarpent:
Yes, or add another signature ;)

PR #547, opened 2022-08-30
Update FindBoost.cmake from upstream

wxmerkt:
Better fix than #546 as per discussion with @nim65s 

Updated from
https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake


nim65s (approved):
Thanks ! This will allow us to use `Boost_NO_WARN_NEW_VERSIONS` to get rid of those annoying warnings :)

wxmerkt:
With some extra fixes to the FindBoost.cmake from upstream, this appears to work for the configurations from 18.04 to 22.04 tested in EigenPy: https://github.com/stack-of-tasks/eigenpy/pull/314

nim65s:
Impressive, thanks a lot for this !

jcarpent (approved):

PR #552, opened 2022-09-09
Release: edit version number in pyproject.toml

fabinsch:
Automatically set the version number in the `pyproject.toml` file, similar to what's done for the `package.xml` . This is useful for projects like [proxsuite](https://github.com/simple-robotics/proxsuite) that use [cmeel](https://github.com/cmake-wheel/cmeel).


nim65s (approved):
Thanks !

PR #554, opened 2022-09-11
Automatically uninstall before install

gergondet:
This PR automatically calls `uninstall` before an install command is called.

- The motivation is to avoid keeping stale libraries and obsolete header files around
- This relies on the documented fact that `install` commands are run in the specified order, hence the "hook" is setup early, normally before any call to `install` in user code happened (i.e. as long as you don't have `install` commands before `include(base.cmake)` it should work
- it changes the behavior of the uninstall target so that it doesn't print an error message if called before install (for the first time)

Since it's an impactful change I will wait for feedback (@nim65s @jcarpent @wxmerkt in particular) before merging


jcarpent (approved):

nim65s (approved):

wxmerkt (commented):
This should be fine - I haven't tested it and/but I dont expect issues with it.

gergondet:
Thanks for the feedback, I'll merge this now.

PR #561, opened 2022-10-06
hpp/idl: remove useless import

nim65s:
Hi,

Trying to release hpp-corbaserver, I got the following error:

```
omniidl: Could not import back-end 'omniidl_be_python_with_docstring'
omniidl: Maybe you need to use the -p option?
omniidl: (The error was 'No module named 'omniidl_be.omniidl_be_python_with_docstring'')
```

I have no idea why, but removing an unused import fixed the issue.

@jmirabel : you added this file in #194, so if you have a better understanding of the issue, or have a better idea, that's welcome. 
Otherwise, I think this change couldn't harm anyway, so we can just merge this and don't loose any time here.


nim65s:
Well, flake8 says I'm wrong, and "comment" is used line 115. In other places we have `comment = node`, should we use this in l.115 too ?

PR #565, opened 2022-11-16
Make FindqpOASES compatible with other distros than linux

ymontmarin:
Make FindqpOASES compatible with other distros than linux

Adding other names with the other extensions possibles.

Best,
Yann

### DIFF
--- a/find-external/qpOASES/FindqpOASES.cmake
+++ b/find-external/qpOASES/FindqpOASES.cmake
@@ -28,7 +28,7 @@ find_path(
   PATHS ${qpOASES_PREFIX} ${qpOASES_PREFIX}/include)
 find_library(
   qpOASES_LIBRARY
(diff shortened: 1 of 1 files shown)


find-external/qpOASES/FindqpOASES.cmake
```diff
@@ -28,7 +28,7 @@ find_path(
   PATHS ${qpOASES_PREFIX} ${qpOASES_PREFIX}/include)
 find_library(
   qpOASES_LIBRARY
-  NAMES libqpOASES.so
+  NAMES libqpOASES.so libqpOASES.dylib qpOASES.lib
```
nim65s, line 31: I guess we should use a name without suffix or prefix.  Could you try with only `qpOASES` ?

nim65s:
Thanks :)

jcarpent (approved):

PR #572, opened 2022-12-21
Fix binary path for logging

jcarpent:


jcarpent:
I will merge it to deploy it on ProxQP.

PR #602, opened 2023-05-02
parse package.xml in python, fix #601

### LINKED ISSUE #601: Cannot COMPUTE_VERSION_FROM_ROS_PACKAGE_XML_FILE on Windows
The following commands are not supported on Windows:
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/7cf918642ffff34e6960b240aae87b0414c28860/version.cmake#L124-L127

nim65s:


nim65s:
`PYTHON_EXECUTABLE` is not available at this point…

alxbilger:
Indeed, `${PYTHON_EXECUTABLE}` is empty

jcarpent:
In general, this is not a good idea to rely on Python as some projects may not need it.

PR #603, opened 2023-05-02
parse package.xml with regex, fix #601

### LINKED ISSUE #601: Cannot COMPUTE_VERSION_FROM_ROS_PACKAGE_XML_FILE on Windows

nim65s:
That's even worse than #602, but might actually work


alxbilger (approved):
I confirm that #601 is fixed with these changes

gergondet (approved):
> That's even worse than https://github.com/jrl-umi3218/jrl-cmakemodules/pull/602, but might actually work

It has the advantage not to rely on anything other than CMake :-)

PR #608, opened 2023-05-31
Refactor python finders and helpers

jcarpent:
I've split the initial `python.cmake` into two parts and extended the PYTHON_BUILD macro to also account for the current config (needed on Windows).


nim65s (commented):
PYTHON_BUILD & PYTHON_BUILD_FILE share most of their code, so I'd suggest to either make one use the other, or define a third macro which would be used by both.

jcarpent:
Thanks @nim65s for the quick review. This is done.

nim65s (approved):

PR #611, opened 2023-06-02
python: include python-helpers

nim65s:
#608 moved some things from `python.cmake` to `python-helpers.cmake`.

As a result, projects updating their submodule are now broken with eg.
```
CMake Error at python/CMakeLists.txt:13 (python_install_on_site):
  Unknown CMake command "python_install_on_site".
```

Here is a quick fix, unless I misunderstood something.


nim65s:
Now, another issue is that the new version of python.cmake without some macros is shipped with the latest eigenpy release, and multiple other packages were relying on it x)

PR #612, opened 2023-06-03
Fix interface issues for cmake>=3.16 && cmake <3.19

jcarpent:


jcarpent:
cc: @fabinsch This PR should solve cmake interface issues on Pinocchio 3x.

PR #620, opened 2023-09-16
require CMake >= 3.10, fix #338

nim65s:
Hi,

[CMake 3.27](https://cmake.org/cmake/help/latest/release/3.27.html#deprecated-and-removed-features) now show the following warning:
```
CMake Deprecation Warning at CMakeLists.txt:27 (cmake_minimum_required):
Compatibility with CMake < 3.5 will be removed from a future version of
CMake.

Update the VERSION argument <min> value or use a ...<max> suffix to tell
CMake that the project does not need compatibility with older versions.
```

So I guess this is time to enforce #338, where 3.10 was the latest consensual version in the discussion.

PR #621, opened 2023-09-19
Apply CMake >= 3.10 policies

nim65s:
ref. #620


nim65s:
@gergondet : I think this version cover all the case we discussed about, but I'm not sure anymore. Do you see anything else ?

gergondet (approved):

PR #624, opened 2023-10-04
python: remove distutils use, fix #623

### LINKED ISSUE #623: Python 3.12 support
We need to remove the use of `distutils` to properly support Python 3.12.
Please see this [note](https://peps.python.org/pep-0632/) related to PEP 632.

This issue triggers an error on conda-forge for EigenPy: https://github.com/conda-forge/eigenpy-feedstock/pull/115

nim65s:
Hi,

This allow compatibility with python 3.12, remove warnings on python 3.10 & 3.11, but break compatibility with python < 3.4 (for pathlib)

NB: I guess this would require multilple tests in different setups to check that this is not breaking anything.

### DIFF
--- a/python.cmake
+++ b/python.cmake
@@ -137,9 +137,8 @@ macro(FINDPYTHON)
       # Provide some hints according to the current PYTHON_EXECUTABLE
       if(NOT DEFINED PYTHON_INCLUDE_DIR)
         execute_process(
-          COMMAND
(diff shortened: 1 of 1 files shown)


nim65s:
Also I didn't change anything in the cython/ directory. 
@gergondet : tell me if you prefer me to handle this in this PR, or if we should do otherwise.

PR #627, opened 2023-10-18
Emit a clear error when a component is not found in a required package

gergondet:
Prior to this PR a call to `find_package(MyProject REQUIRED COMPONENTS MyComponent)` would fail.

For example: (here `find_package(baseline_walking_controller REQUIRED COMPONENTS FootstepPlanner)`)
```
-- baseline_walking_controller: FootstepPlanner not found.
-- C++ standard sufficient: Minimal required 11, currently defined: 14
CMake Error at CMakeLists.txt:5 (find_package):
  Found package configuration file:

    /usr/local/lib/cmake/baseline_walking_controller/baseline_walking_controllerConfig.cmake

  but it set baseline_walking_controller_FOUND to FALSE so package
  "baseline_walking_controller" is considered to be NOT FOUND.
  ```

The reason it fails is not easy to scan (here it is the top line but the real output there's many messages before that one)

With this PR we get the following error instead:
```
CMake Error at /usr/local/lib/cmake/baseline_walking_controller/baseline_walking_controllerConfig.cmake:180 (message):
  baseline_walking_controller: FootstepPlanner not found.
  CMakeLists.txt:5 (find_package)
```


nim65s (approved):
nice, thanks !

PR #634, opened 2023-11-11
remove print message

jcarpent:


gergondet:
I think @nim65s has integrated this in #633

PR #637, opened 2023-11-20
Update PATH when running generate_stubs on Windows

jorisv:
`generate_stubs` need to load an uninstalled Python binding. 
This can be an issue when the Python binding have internal dependencies on Windows.
This PR modify the PATH to allow Python to find those internal dependencies.

On Linux and Mac, this is not an issue thanks to the RPATH.

### DIFF
--- a/stubgen/CMakeLists.txt.in
+++ b/stubgen/CMakeLists.txt.in
@@ -4,7 +4,7 @@ project(stubgen NONE)
 
 include(ExternalProject)
 ExternalProject_Add(stubgen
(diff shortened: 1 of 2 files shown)


nim65s (commented):
I can't really help reviewing this, as my knowledge with WIN32 is null, sorry.

jorisv:
> I can't really help reviewing this, as my knowledge with WIN32 is null, sorry.

Ok, no problem, you can remove yourself from the review (I didn't find if I can do it myself).

PR #638, opened 2023-11-20
Add ADD_PYTHON_MEMORYCHECK_UNIT_TEST macro

jorisv:
ADD_PYTHON_MEMORYCHECK_UNIT_TEST macro run Python unit tests through valgrind to detect memory leaks.

### DIFF
--- a/test.cmake
+++ b/test.cmake
@@ -117,14 +117,14 @@ endmacro(
   NAME
   SOURCE)
 
-# .rst: .. command:: COMPUTE_PYTHONPATH (result [MODULES...])
+# .rst: .. command:: COMPUTE_PYTHONPATH ([MODULES...])
 #
 # Get the `PYTHONPATH` environment variable with each `MODULES` appended to it
 # as: `CMAKE_BINARY_DIR/MODULE_PATH`
 #
 # .. note:: :command:`FINDPYTHON` should have been called first.
 #
-function(COMPUTE_PYTHONPATH result)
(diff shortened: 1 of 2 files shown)


jcarpent (approved):

test.cmake
```diff
@@ -172,12 +157,69 @@ macro(ADD_PYTHON_UNIT_TEST NAME SOURCE)
...
+    set(PYTHON_TEST_SCRIPT "${PROJECT_SOURCE_DIR}/${SOURCE}")
+    configure_file(
+      ${PROJECT_JRL_CMAKE_MODULE_DIR}/memorycheck_unit_test.cmake.in
+      ${TEST_FILE_NAME} @ONLY)
+
+    add_test(NAME ${NAME} COMMAND ${CMAKE_COMMAND} -P ${TEST_FILE_NAME})
+
+    set(MODULES "${ARGN}") # ARGN is not a variable
+    compute_pythonpath(PYTHONPATH ${MODULES})
+    set_tests_properties(${NAME} PROPERTIES ENVIRONMENT "${ENV_VARIABLES}")
```
nim65s, line 219: `ENV_VARIABLES` is not defined here, as it stays in the scope of `compute_pythonpath` because it is a function and not a  macro. I'll propose a fix in a commit.

nim65s (approved):
I'm not sure `ARGN` processing is perfect, but it looks like this is working well enough in my test cases. Thanks !

PR #641, opened 2023-11-22
version: fix out-of-tree builds

### LINKED ISSUE #640: can't get version number when building outside of the source tree
As reported in https://github.com/stack-of-tasks/eigenpy/issues/405:
```
git clone https://github.com/stack-of-tasks/eigenpy
cmake -B build -S eigenpy
```
raise:
```
-- JRL cmakemodules not found. Let's fetch it.
-- Configuring "eigenpy" (http://github.com/stack-of-tasks/eigenpy)
fatal : ni ceci ni aucun de ses répertoires parents (jusqu'au point de montage /) n'est un dépôt git
Arrêt à la limite du système de fichiers (GIT_DISCOVERY_ACROSS_FILESYSTEM n'est pas défini).
```

and then the project version number is 0.0.0

I guess this WORKING_DIRECTORY is not working here
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/0b7f479df6f3825d48eeaaf4f6ba57f80fa51cff/version.cmake#L62

Because, according to the docs:
> This is the source directory of the last call to the `project()`

And we didn't run `project()` yet

nim65s:
eg. when module is fetched instead of cloned
fix #640
ref https://github.com/stack-of-tasks/eigenpy/issues/405

Not sure if we should keep the previous behavior with `"${PROJECT_JRL_CMAKE_MODULE_DIR}/.."` or not.


nim65s:
This PR is correctly fixing the regression test added in https://github.com/stack-of-tasks/eigenpy/pull/406

jcarpent (approved):

PR #652, opened 2024-01-17
Add a macro to do explicit template instantiation on Windows/MacOS/Linux

jorisv:
Explicit template instantiation is done with the `extern template` keyword.

Unfortunately, when used with `dllexport` on Windows this create a warning message.

To avoid that, we define `@LIBRARY_NAME@_EXP_INST_DECL` that manage the compiler and the import/export context to avoid warnings.

### DIFF
--- a/config.hh.cmake
+++ b/config.hh.cmake
@@ -31,37 +31,53 @@
 //
 // On Linux, set the visibility accordingly. If C++ symbol visibility
 // is handled by the compiler, see: http://gcc.gnu.org/wiki/Visibility
+//
+// Also define explicit template instantiation keyword
+// This keyword should be inline template but when mixed with
+// dllexport on Windows this generate a warning.
+// The inline keyword is then remove when mixed with dllexport.
 # if defined _WIN32 || defined __CYGWIN__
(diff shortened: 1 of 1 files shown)


config.hh.cmake
```diff
@@ -31,37 +31,53 @@
...
+// Also define explicit template instantiation keyword
+// This keyword should be extern template but when mixed with
+// dllexport on Windows this generate a warning.
+// The extern keyword is then remove when mixed with dllexport.
 # if defined _WIN32 || defined __CYGWIN__
 // On Microsoft Windows, use dllimport and dllexport to tag symbols.
 #  define @LIBRARY_NAME@_DLLIMPORT __declspec(dllimport)
 #  define @LIBRARY_NAME@_DLLEXPORT __declspec(dllexport)
 #  define @LIBRARY_NAME@_DLLLOCAL
+#  define @LIBRARY_NAME@_EXP_INST_DECL_IMPORT extern template
```
jcarpent, line 44:
```suggestion
#  define @LIBRARY_NAME@_EXPLICIT_INSTANTIATION_IMPORT extern template
```

PR #653, opened 2024-01-17
Use add_test new syntax

jorisv:
`add_test` new syntax can:

- Take a target as argument (no need to provide the full path)
- Target properties can use generator expression


jcarpent:
Is it applicable to all CMake versions starting from 3.10?

jorisv:
Yes : https://cmake.org/cmake/help/v3.10/command/add_test.html#add-test

PR #659, opened 2024-01-26
Add CXX_FLAGS_BY_COMPILER_FRONTEND function

jorisv:
This function allow to choose different compiler options or definitions based on the compiler frontend

### DIFF
--- a/compiler.cmake
+++ b/compiler.cmake
@@ -72,7 +72,8 @@ endmacro(_SETUP_PROJECT_WARNINGS)
 .. command:: CXX_FLAGS_BY_COMPILER_FRONTEND(<CLANG [<flags1>...]>
                                             <GCC   [<flags1>...]>
                                             <MSVC  [<flags1>...]>
-                                            OUTPUT flags)
+                                            OUTPUT flags
+                                            <FILTER>)
 
 Detect the compiler frontend (the command line interface) and output the
 corresponding ``CXX_FLAGS``.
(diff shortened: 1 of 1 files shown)


jorisv:
We can maybe use this macro in _SETUP_PROJECT_WARNINGS

compiler.cmake
```diff
@@ -19,51 +19,140 @@ macro(_SETUP_PROJECT_WARNINGS)
...
+
+function(CXX_FLAGS_BY_COMPILER_FRONTEND)
+  set(options FILTER)
+  set(oneValueArgs OUTPUT)
+  set(multiValueArgs GNU MSVC)
+  cmake_parse_arguments(ARGS "${options}" "${oneValueArgs}" "${multiValueArgs}"
+                        ${ARGN})
+
+  # Before CMake 3.14, the CMAKE_CXX_COMPILER_FRONTEND_VARIANT doesn't exists.
+  # Before CMake 3.26, the CMAKE_CXX_COMPILER_FRONTEND_VARIANT doesn't exists
```
nim65s, line 115: There is something strange with those 2 comments

nim65s (approved):
Looks good to me, except for a misleading comment

compiler.cmake
jorisv: The comment line 115 continue line 116. The cmake-format has split the line.

I can rewrite it to make it clearer.

PR #694, opened 2024-05-13
Add package.xml

MaximilienNaveau:
Add a package.xml file so one can use this package as a dependency directly.
I added a list of maintainer in this file.
It's hard to decide in practice as this project has many maintainers and contributors.
So I picked 3 names among the most commiters according to github...
If it's not politically correct tell me I could:
- remove the list altogether and let the people contact the maintainers from the github utils.
- add people in this list
- remove people in this list
- others?

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -5,7 +5,7 @@
   <description>CMake utility toolbox</description>
 
   <maintainer email="justin.carpentier@inria.fr">Justin Carpentier</maintainer>
-  <maintainer email="gsaurel@laas.fr">Guilhem Saurel</maintainer>
+  <maintainer email="guilhem.saurel@laas.fr">Guilhem Saurel</maintainer>
   <maintainer email="fill.me@in.com">Pierre Gergondet</maintainer>
 
 
--- a/package.xml
+++ b/package.xml
@@ -4,10 +4,9 @@
   <version>0.0.0</version>
   <description>CMake utility toolbox</description>
 
-  <maintainer email="justin.carpentier@inria.fr">Justin Carpentier</maintainer>
   <maintainer email="guilhem.saurel@laas.fr">Guilhem Saurel</maintainer>
-  <maintainer email="fill.me@in.com">Pierre Gergondet</maintainer>
-
+  <author>Justin Carpentier</author>
+  <author>Pierre Gergondet</author>
 
   <license>BSD-2</license>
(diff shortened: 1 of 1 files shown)


package.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>jrl-cmakemodules</name>
+  <version>0.0.0</version>
+  <description>CMake utility toolbox</description>
+
+  <maintainer email="justin.carpentier@inria.fr">Justin Carpentier</maintainer>
+  <maintainer email="gsaurel@laas.fr">Guilhem Saurel</maintainer>
+  <maintainer email="fill.me@in.com">Pierre Gergondet</maintainer>
```
nim65s, line 9:
```suggestion
  <maintainer>Pierre Gergondet</maintainer>
```

package.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>jrl-cmakemodules</name>
+  <version>0.0.0</version>
+  <description>CMake utility toolbox</description>
+
+  <maintainer email="justin.carpentier@inria.fr">Justin Carpentier</maintainer>
+  <maintainer email="gsaurel@laas.fr">Guilhem Saurel</maintainer>
```
nim65s, line 8:
```suggestion
  <maintainer email="guilhem.saurel@laas.fr">Guilhem Saurel</maintainer>
```

nim65s (approved):
This maintainer list is mostly used by ROS buildfarm, so as long as we don't publish this inside https://github.com/ros/rosdistro/ this list shouldn't hurt anyone.

Still, yes, it could be good to define a proper list of authors / maintainers in this package in general.

package.xml
MaximilienNaveau: Unfortunately I cannot as the xml note maintainer asks for an email.

nim65s: I guess this is because this is mostly for use by ROS buildfarm, to spam people responsible for broken packages. I think nobody want that. Can we use an "author" node instead, then ?

MaximilienNaveau: Hi I need I maintainer in the package.xml and I can put the rest as authors.
Can I put you @nim65s as sole maintainer?
And the rest as author?
So there is on email that is being spammed?

nim65s: yes

PR #717, opened 2024-10-04
ROS 2 file generation macro

### LINKED ISSUE #716: ROS 2 file generation macro
Dear @jrl-cmakemodules maintainers,
Thanks for the great work.
They are many packages using jrl-cmakemodules which can be detected by the new ament package system simply by generating three files. 
In pinocchio this is done [here](https://github.com/stack-of-tasks/pinocchio/blob/0b594a0fb803fcc2141856e6fb8c69bcf1e8981b/CMakeLists.txt#L512)

The 10 corresponding lines were happily copy in various other repositories (see [master_board_sdk for instance](https://github.com/open-dynamic-robot-initiative/master-board/blob/89ca74da0b515d3b939051dfe6b69b4cd6514877/sdk/master_board_sdk/CMakeLists.txt#L178))

It would be nice to have a corresponding macro called in SETUP_PROJECT_FINALIZE when for instance the ROS_VERSION environment variable is set to 2 or when a ros2.cmake file is included.

What are your thoughts on this proposal ?

olivier-stasse:
This is a proposal to fix #716.
This PR introduces a new file ros2.cmake.
It provides a macro _install_project_ros2_ament_files to generate files for the ROS2 ament system.
This macro is called in setup_project_finalize() in base.cmake.
The generation works if the variable BUILDING_ROS2_PACKAGE is set (to TRUE
for instance).

pre-commit was run over the PR.
Tested against unittest.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,7 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
           ./release.cmake
           ./relpath.cmake
           ./ros.cmake
+          ./ros2.cmake          
           ./sdformat.cmake
           ./setup.cfg
           ./shared-library.cmake
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,7 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
           ./release.cmake
           ./relpath.cmake
           ./ros.cmake
-          ./ros2.cmake          
+          ./ros2.cmake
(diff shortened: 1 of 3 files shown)


nim65s (changes requested):
Can you add this new `ros2.cmake` file in `CMakeLists.txt` ?

olivier-stasse:
> Can you add this new `ros2.cmake` file in `CMakeLists.txt` ?

Done. Thanks for the review @nim65s !

nim65s (approved):

PR #720, opened 2024-10-07
Add FindIpopt.cmake module

ManifoldFR:
This PR:
- adds a FindIpopt.cmake module, which enables the use of `find_package(Ipopt)`

### DIFF
--- a/find-external/Ipopt/FindIpopt.cmake
+++ b/find-external/Ipopt/FindIpopt.cmake
@@ -28,18 +28,29 @@ else()
   message(STATUS "  Ipopt library dirs: ${Ipopt_LIBRARY_DIRS}")
   message(STATUS "  Ipopt include dirs: ${Ipopt_INCLUDE_DIRS}")
   add_library(ipopt SHARED IMPORTED)
+  # On Windows, ipopt library is named ipopt-3
   find_library(
     ipopt_lib_path
-    NAMES ipopt
-    PATHS ${Ipopt_LIBRARY_DIRS})
+    NAMES ipopt ipopt-3
+    PATHS ${Ipopt_LIBRARY_DIRS} REQUIRED)
   message(STATUS "  Ipopt library found at ${ipopt_lib_path}")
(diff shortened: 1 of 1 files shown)


find-external/Ipopt/FindIpopt.cmake
```diff
@@ -0,0 +1,48 @@
+# Copyright 2024 CNRS INRIA
+#
+# Author: Wilson Jallet
+#
+# Adapted from: https://github.com/casadi/casadi/blob/main/cmake/FindIPOPT.cmake
```
nim65s, line 5: I think this file should keep the original LGPL-3.0 license then

nim65s (approved):

ManifoldFR:
Hi folks, are we merging this after all ? Do you have feedback @jorisv ?

jorisv (changes requested):
I will test FindIpopt on Linux and Windows before approving

find-external/Ipopt/FindIpopt.cmake
```diff
@@ -0,0 +1,56 @@
...
+  message(STATUS "  Ipopt library found at ${ipopt_lib_path}")
+  set_target_properties(
+    ipopt
+    PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${Ipopt_INCLUDE_DIRS}"
+               INTERFACE_COMPILE_OPTIONS "${Ipopt_CFLAGS_OTHER}"
+               IMPORTED_CONFIGURATIONS "RELEASE")
+
+  if(NOT CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+    target_compile_definitions(ipopt INTERFACE HAVE_CSTDDEF)
+  endif()
```
jorisv, line 45: I don't know if this workaround is still needed : https://github.com/ghorn/debian-coinor-ipopt/issues/2

I test without it and it's working on conda-forge Ipopt.
Maybe it's mandatory for older Ipopt.

ManifoldFR: Okay, perhaps we could remove it and ask for a given minimum version if Ipopt?

jorisv (approved):

PR #723, opened 2024-10-29
Remove Windows path separator in PYTHON_SITELIB

jorisv:
When PYTHON_SITELIB is not provided and is searched from python interpreter call, the path in not normalized and can contains `\`.

This create an error when calling `python_install_on_site` macro.


jorisv:
Tested in this PR: https://github.com/Simple-Robotics/proxsuite-nlp/pull/113

python.cmake
```diff
@@ -351,6 +349,11 @@ macro(FINDPYTHON)
     endif()
   endif()
 
+  # Avoid Windows path
```
olivier-roussel, line 352: This comment confused me a bit. Maybe something a bit more explicit like:
`# Avoid paths in Windows format`

olivier-roussel (approved):

PR #729, opened 2024-12-15
Add normalization step for install_path

jcarpent:
Related to new CMake policy 0177 since CMake >= 3.31. See https://cmake.org/cmake/help/latest/policy/CMP0177.html

### DIFF
--- a/header.cmake
+++ b/header.cmake
@@ -1,4 +1,4 @@
-# Copyright (C) 2008-2019 LAAS-CNRS, JRL AIST-CNRS, INRIA.
+# Copyright (C) 2008-2024 LAAS-CNRS, JRL AIST-CNRS, INRIA.
 #
 # This program is free software: you can redistribute it and/or modify it under
 # the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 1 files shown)


header.cmake
```diff
@@ -316,9 +316,10 @@ macro(HEADER_INSTALL)
     string(REGEX REPLACE "${CMAKE_BINARY_DIR}" "" DIR "${DIR}")
     string(REGEX REPLACE "${PROJECT_SOURCE_DIR}" "" DIR "${DIR}")
     string(REGEX REPLACE "include(/|$)" "" DIR "${DIR}")
+    cmake_path(SET INSTALL_PATH NORMALIZE "${CMAKE_INSTALL_INCLUDEDIR}/${DIR}")
```
nim65s, line 319:
```suggestion
    if(CMAKE_VERSION` VERSION_GREATER 3.20)
      # workaround CMP0177
      cmake_path(SET INSTALL_PATH NORMALIZE "${CMAKE_INSTALL_INCLUDEDIR}/${DIR}")
    else()
      set(INSTALL_PATH "${CMAKE_INSTALL_INCLUDEDIR}/${DIR}")
    endif()
```

As https://cmake.org/cmake/help/latest/command/cmake_path.html#command:cmake_path was added in 3.20, and we don't really fear breaking things for versions < 3.20

nim65s (changes requested):

PR #736, opened 2025-01-13
install ament files by default

nim65s:
ref. https://github.com/Gepetto/example-robot-data/pull/261 / https://github.com/stack-of-tasks/pinocchio/pull/2554

### DIFF
--- a/ros2.cmake
+++ b/ros2.cmake
@@ -13,6 +13,8 @@
 # You should have received a copy of the GNU General Public License along with
 # this program.  If not, see <http://www.gnu.org/licenses/>.
 
+option(BUILDING_ROS2_PACKAGE "Install ament specific files" ON)
+
 # This macro creates 3 files which are allowing a package to be found using the
 # AMENT system present in ros 2 Setting BUILDING_ROS2_PACKAGE is enough to
 # trigger this behavior. _install_project_ros2_ament_files()


jorisv:
I'm got mixed feeling about this PR.

This will install this file in all our packages but this is only useful for ROS2 build system.
Did `colcon` setup some CMake variables to be able to recognize when we need to create ament_index file ?

nim65s:
I don't understand your point.
Do you think a package installed from aur / brew / conda / nix / pip / robotpkg, should not be available for use in a ros workspace ?

Yes, a pair of useless files (one empty, and one with a single line) will be installed, but how big of an issue is this ? We are already installing package.xml all the time. And we have been installing those files in pinocchio for some time.

And for people who really don't want those 2 extra files, an opt-out option is still available.

jorisv:
I'm not an expert in ROS workspace.
I was thinking this file is only mandatory when building all our stack from source with colcon.

Also, is-it safe to mix binary package from aur / brew / conda /nix / pip / robotpkg with a ROS workspace ?
If it's a valid use case I don't have any problem with that.

nim65s:
I think those files are used at runtime by ros tooling. So far we had no issues in our packages build by colcon without those files.

And yes, some mixes are unsafe, but there are plenty of safe ones. With aur / robotpkg, there are no unsafe options. Also, packages without binaries (eg. example-robot-data) are always safe.

jorisv (approved):

PR #738, opened 2025-01-15
Doxygen: simplify logic, remove vendored header/footer/css files

### LINKED ISSUE #737: Updating `doxygen.cmake` behaviour (remove old files?)
I am opening this issue to discuss how to update and simplify our Doxygen module so it behaves in a more predictable way and gets a bit less in the way of the user.

One thing I've encountered is strange behaviour with the `DOXYGEN_USE_TEMPLATE_CSS` option, and I've gotten burned with some of the outdated template doxygen.css, header.html, and footer.html included in the repo.

**When `DOXYGEN_USE_TEMPLATE_CSS` is undefined or `False`**

If the variable `CMAKE_HTML_{STYLESHEET,HEADER,FOOTER}` is undefined, `doxygen.cmake` points `HTML_{STYLESHEET,HEADER,FOOTER}` to previously generated template files located [in](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/doxygen/doxygen.css) [this](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/doxygen/header.html) [repository](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/doxygen/footer.html) - last updated 3, 6 and 10 years ago respectively.

For stylesheet and footer, this is done in [this part](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/6b0564f45af29a90160aafdfd67eb7a07ace48ed/doxygen.cmake#L595-L602) of the code.
For the header, it is done [somewhere else](https://github.com/jrl-umi3218/jrl-cmakemodules/blob/6b0564f45af29a90160aafdfd67eb7a07ace48ed/doxygen.cmake#L744-L747).

A funny caveat in the behaviour:
- If the variable is an **empty string**, it is considered set to the empty string and `doxygen.cmake` does _not_ change them. Then the behaviour is Doxygen's default behaviour when these config values are empty, which is to use _up-to-date_ template files shipped with Doxygen.
- If the variable is _unset_ (e.g. in CMake via `set(DOXYGEN_HTML_HEADER)`) then we use the (super ancient) file shipped with jrl-cmakemodules.

**When `DOXYGEN_USE_TEMPLATE_CSS` **is** defined and `True`**

A set of _up-to-date_ standard header and footer are generated and put in the doc build-tree, our module executes `doxygen -w html` as described on [this page](https://www.doxygen.nl/manual/doxygen_usage.html). Then:
- if the CMake variable `DOXYGEN_HTML_{FOOTER,HEADER,STYLESHEET}` are unset, it uses the corresponding build-tree file
- if the CMake variable is set, then that value is used

To summarize, `DOXYGEN_USE_TEMPLATE_CSS` does behaves as follows for each of the files (css stylesheet, html footer, html header): 

| `DOXYGEN_USE_TEMPLATE_CSS` | False/undefined | True |
| -| -| -|
| `DOXYGEN_{STYLESHEET,HEADER,FOOTER}` undefined | use ancient files included in this repo | generated up-to-date standard file in build-tree, use it (**end result same as Doxygen default**) |
| empty string `''` | use (up-to-date) standard file  shipped with Doxygen (default) | use Doxygen default |
| some `path/to/file` | use specified file | use specified file |

**Weird look with old template files**

The old stylesheet and html footer and header can create poor looking docs.  
Since Doxygen 1.8.19 the Doxygen logo is an svg file (the footer.html generated since then takes this into account, changing the old `doxygen.png`). But the file shipped in this repo is so old we get a dead image source:
Screenshot From 2025-01-14 16-05-13
Instead of
Screenshot From 2025-01-14 16-07-34

## Suggested changes

- Remove the `DOXYGEN_USE_TEMPLATE_CSS` option. To me it looks like it could create some unexpected behaviour unless the user specifies `DOXYGEN_{STYLESHEET, HEADER, FOOTER}` to some non-empty string value.
- Remove the shipped generated `footer.html`, `header.html`, and `doxygen.css`: keeping old files in this repo is bad because Doxygen itself might be updated in a way the old versions of these files do not expect. We could have a bot update these files (with `doxygen -w html`) quarterly, but since the validity of these files depends on the version of the Doxygen executable, we should trust the end-user or packaging environment's Doxygen to generate up-to-date files in the build tree.
- We should now be using `HTML_EXTRA_STYLESHEET`, instead of `HTML_STYLESHEET` which is **deprecated.** It is available since Doxygen 1.8.2 (2012), see the [changelog](https://www.doxygen.nl/manual/changelog.html):
  > Added option HTML_EXTRA_STYLESHEET which allows adding an additional stylesheet without completely replacing doxygen.css.

## Other questions

- Is the possibility to set the Doxygen configuration from CMake using the `DOXYGEN_${option}` really useful? From what I see it
  1. requires us to maintain the list of options in the `foreach()` loop in case options are added to/removed from Doxygen
  2. adds a bit of complexity to `doxygen.cmake`
  3. doesn't seem to do more than writing a `Doxygen.extra.in` file ourselves
- If we want to ship a header.html or footer.html, we could maybe offer the option of using KaTeX or mathjax3 which are much faster LaTeX rendering backends for the web.
- We could offer the option to use the much more pleasant [doxygen-awesome](https://github.com/jothepro/doxygen-awesome-css) css files

ManifoldFR:
Following the discussion on my issue #737, here is my proposed change.

I will propose an update to the vendored MathJax in a subsequent PR.

Resolves #737 

cc @nim65s @jmirabel


jmirabel:
Is it supported to use doxygen without providing a header, footer and css ?
In the past, it used not to be the case.

ManifoldFR:
> Is it supported to use doxygen without providing a header, footer and css ? In the past, it used not to be the case.

Yes! From what I'm seeing in the [docs](https://www.doxygen.nl/manual/config.html#cfg_html_header):
> The HTML_HEADER tag can be used to specify a user-defined HTML header file for each generated HTML page. If the tag is left blank Doxygen will generate a standard header.

I'm not aware that it was ever _not_ the case however. I started looking around at some resolved issues on doxygen's repo from like over a decade ago (like https://github.com/doxygen/doxygen/issues/4079), and I'm not finding anything.

jcarpent (approved):

ManifoldFR:
Hello, is this thing fine to merge? I also have a MathJax-upgrade PR ready to go after this 😃

ManifoldFR:
Thanks @nim65s :)

PR #740, opened 2025-01-22
Add support for MathJax 3, upgrade vendored MathJax to 3.2.2

ManifoldFR:
Follow up to #737 and #738

This pull request updates the vendored files for MathJax to the latest release, 3.2.2.

The previous version that was shipped with jrl-cmakemodules was from an old 2.x from over a decade ago. MathJax 3.x is a full rewrite which is much faster than 2.x (partly as a response to the development of KaTeX which was leaner and faster).

## Major changes

- Upgraded vendored version of MathJax to [3.2.2 (June 2022)](https://github.com/mathjax/MathJax/releases/tag/3.2.2)
- Added support for Doxygen option `MATHJAX_VERSION`, which defaults to `MathJax_3` (for Doxygen>=1.9.2. Other users get MathJax 2.)
- In case MathJax version is set to 3, default value of `MATHJAX_RELPATH` is set to 
`${PROJECT_JRL_CMAKE_MODULE_DIR}/doxygen/MathJax` (i.e. our vendored MathJax). If the user desires MathJax 2, they can set the path themselves, or leave it unset to let Doxygen inject the MathJax 2.x CDN instead.
- Added the entirety of `doxygen/MathJax` to the pre-commit exclusion rules.


nim65s (approved):

jcarpent (approved):

ManifoldFR:
@jorisv has found there is still a bug on Doxygen 1.9.1, that I've fixed in the latest commit.

PR #744, opened 2025-01-24
doxygen.cmake : make tagfiles processing more consistent

ManifoldFR:
- process list as a comma-separated native CMake list (less awkward)
- fix extraneous whitespaces
- always use absolute paths (allows dedupe to work)
- remove custom `REMOVE_DUPLICATES()` function

### DIFF
--- a/doxygen.cmake
+++ b/doxygen.cmake
@@ -674,19 +674,6 @@ macro(_SETUP_PROJECT_DOCUMENTATION)
   endif(NOT DOXYGEN_FOUND)
 endmacro(_SETUP_PROJECT_DOCUMENTATION)
 
-# REMOVE_DUPLICATES
-# -----------------
-#
-# Remove duplicate values from a space separated list
-function(REMOVE_DUPLICATES ARG_STR OUTPUT)
-  set(ARG_LIST ${ARG_STR})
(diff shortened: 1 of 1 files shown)


jcarpent (approved):

ManifoldFR:
As discussed with @jorisv, the following lines:
https://github.com/jrl-umi3218/jrl-cmakemodules/blob/4d66e417d3aad73cdad7b4fede6f9815c00c4430/doxygen.cmake#L642-L646

Mess up the installation of doxytags when doing `cmake <insert-cmake-args-here> && make install`.
Because on the first CMake configuration, the doxytag file in `build/doc` doesn't exist yet, so the `install()` for it is not registered.

ManifoldFR:
We've found a workaround with @jorisv, which is to make the doxytag file's build-tree relative location the same as its install-tree location -- that is, `doc/doxygen-html`. Before, the locations would be `<build-dir>/doc/${PROJECT_NAME}.doxytag` and `<install-dir>/doxygen-html/${PROJECT_NAME}.doxytag`.

We gain:
- better consistency
- no need that the tagfile exists before installing (let it be carried with the `doxygen-html` directory which is installed wholesale)
- easier to clean up

PR #745, opened 2025-02-13
add BUILD_STANDALONE_PYTHON_INTERFACE option

nim65s:
Hi,

This is PR is not ready, but we can already start some discussions.

It add the `BUILD_PYTHON_INTERFACE` option we already have in many places, plus a `BUILD_STANDALONE_PYTHON_INTERFACE` one.

The idea of that second one is that, using a single project source, we should be able to package one version without any python interface, then install that, then package another version with only the python interface, linked to the first one with a standard dependency.

The packaging split of those python bindings are already done in robotpkg through some patches there, but would help in many other places, so I think this should be upstreamed.

For example, this would fix https://github.com/stack-of-tasks/pinocchio/issues/2575 because we could have a dependency chain like example-robot-data <- pinocchio <- py-pinocchio <- py-example-robot-data.

It would also allow end users to avoid python dependencies when they only care about our C++ interfaces, since they could apt/brew/conda/nix/pacman/pip install either "pinocchio" or "py-pinocchio" (the later coming with the former as a dependency)

It would also reduce the required resources to build our packages in package managers which offer multiple python versions, since we won't have to rebuild pinocchio-cpp for each pinocchio-py38 .. pinocchio-py313

(obviously, the name of the packages with/without python, eventually with the python version number in the package name will be left to choose for each package manager. eg. I think for pip we should have something like coal-cpp and keep coal for the python bindings, but for robotpkg this is already coal / py312-coal, and for nix this is already coal / python312Packages.coal)

It would also allow support of multiple python versions installed side by side on a system, eg:

<details>
  <summary>Install example of coal with both python 3.12 and 3.13</summary>

```
 /tmp/coal/prefix/merged/
├──  include/
│   └──  coal/
│       └── […]
├──  lib/
│   ├──  cmake/
│   │   └──  coal/
│   │       ├──  coalConfig.cmake
│   │       ├──  coalConfigVersion.cmake
│   │       ├──  coalTargets-relwithdebinfo.cmake
│   │       ├──  coalTargets.cmake
│   │       └──  cxx-standard.cmake
│   ├──  libcoal.so -> libcoal.so.3.0.1
│   ├──  libcoal.so.3.0.1
│   ├──  pkgconfig/
│   │   └──  coal.pc
│   ├──  python3.12/
│   │   └──  site-packages/
│   │       └──  coal/
│   │           ├──  __init__.py
│   │           ├──  coal_pywrap.cpython-312-x86_64-linux-gnu.so
│   │           ├──  viewer.py
│   │           └──  windows_dll_manager.py
│   └──  python3.13/
│       └──  site-packages/
│           └──  coal/
│               ├──  __init__.py
│               ├──  coal_pywrap.cpython-313-x86_64-linux-gnu.so
│               ├──  viewer.py
│               └──  windows_dll_manager.py
└──  share/
    ├──  ament_index/
    │   └──  resource_index/
    │       └──  packages/
    │           └──  coal
    └──  coal/
        ├──  hook/
        │   ├──  ament_prefix_path.dsv
        │   └──  python_path.dsv  # NB: this won't work with ament
        └── 󰗀 package.xml
```
</details>

PS: in HPP, we already have an early version of this feature under the `INSTALL_PYTHON_INTERFACE_ONLY` name, but this lacks support of various things from jrl-cmakemodules, like not installing the headers / cmake exports / ros stuff


jorisv (changes requested):
I will review it as soon as I can

nim65s:
rebased

post-project.cmake
```diff
@@ -47,8 +47,36 @@ cmake_dependent_option(
   BUILD_DOCUMENTATION
   OFF
 )
-option(INSTALL_GENERATED_HEADERS "Generate and install standard headers" ON)
-option(INSTALL_PKG_CONFIG_FILE "Generate and install standard .pc file" ON)
+
+option(BUILD_PYTHON_INTERFACE "Build the python bindings" ON)
+cmake_dependent_option(
+  BUILD_ONLY_PYTHON_INTERFACE
```
cmastalli, line 53: As discussed in another thread, I would suggest renaming these variables to `BUILD_PYTHON_INTERFACE_ONLY`. The reason is that the current name is slightly awkward.

nim65s: Thanks for raising this. We discussed this and `BUILD_STANDALONE_PYTHON_INTERFACE` seems more relevant. Would you agree with this naming @cmastalli ?

cmastalli: I would!

jorisv (approved):

PR #757, opened 2025-04-07
Add FFmpeg finder module

### LINKED ISSUE #754: Finder module for FFmpeg
Hello,

Candlewick, which has support for recording videos through FFmpeg, has its own [finder for the FFmpeg libraries](https://github.com/Simple-Robotics/candlewick/blob/main/modules/FindFFmpeg.cmake).

I think I should upstream it here, most likely with some enhancements. For instance it relies on pkg-config which we should rather be sunsetting in the future.

Some other projects provide their own finders for FFmpeg:
- https://github.com/snikulov/cmake-modules/blob/master/FindFFmpeg.cmake
- [Kitware's VTK](https://github.com/Kitware/VTK/blob/master/CMake/FindFFMPEG.cmake)

Would that be interesting? What do you think @nim65s?

ManifoldFR:
Resolves #754

### DIFF
--- a/find-external/FFmpeg/FindFFmpeg.cmake
+++ b/find-external/FFmpeg/FindFFmpeg.cmake
@@ -23,13 +23,11 @@ string(REPLACE ";" " " _ffmpeg_components "${FFmpeg_FIND_COMPONENTS}")
 message(STATUS "Looking for FFmpeg with components ${_ffmpeg_components}")
 unset(_ffmpeg_components)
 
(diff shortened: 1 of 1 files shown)


ManifoldFR:
I've rebased on master after #755, #756 were merged.

ManifoldFR:
@nim65s you think this could be merged?

jorisv (approved):

PR #759, opened 2025-04-11
Tells Clang to pass -fopenmp to the preprocessor rather than frontend

cmastalli:
Hi folks,

I believe we should use -Xpreprocessor rather than -Xclang. Please let me know your thoughts. I am unaware of the reasons why cmake is -Xclang.

Thanks,


cmastalli:
Any feedback regarding this PR @nim65s?

PR #761, opened 2025-05-02
Allow to use some Python-related macros without calling our `FINDPYTHON`

ManifoldFR:
This PR removes the assumption that the CMake variable `PYTHON_EXECUTABLE` is defined (which is done when `FINDPYTHON` is called), replacing it with the assumption that `Python_EXECUTABLE` is defined.

This latter variable is defined in the following scenarios:
- `FINDPYTHON` is called (one branch was missing, I've fixed it)
- the user uses the modern `find_package(Python [...])` from the modern [`FindPython.cmake`](https://cmake.org/cmake/help/latest/module/FindPython.html) module


jcarpent (approved):

nim65s (approved):

jorisv (changes requested):
I will check it soon

ManifoldFR:
Following discussion with @jorisv, this PR will be closed because it would perhaps be too large a change. Jrl-cmakemodules v2 will be when we start breaking things (and then, we'd remove the godforsaken `python.cmake` altogether).

PR #764, opened 2025-05-20
cmake: bump version to 3.22

gwennlbh:


nim65s (approved):
Thanks !

(context for others: this is the version of cmake in Ubuntu 22.04, which is the oldest thing we support since 20.04 is EOL)

nim65s:
Also ref #620

PR #767, opened 2025-06-23
package.xml: clarify license

### LINKED ISSUE #763: Licenses clarifications
Quoting from https://github.com/ros/rosdistro/pull/45744#issuecomment-2892703393:

>     * [ ]  License correctly listed in package.xmls
>       
>       * BSD-3 is correctly listed: https://github.com/Simple-Robotics/nanoeigenpy/blob/e02383cc9f56fefde84273a4a07972c92a6491fa/package.xml#L9
>       * However, the repo/package has a submodule under the `cmake/` directory that has LGPL-licensed code: https://github.com/jrl-umi3218/jrl-cmakemodules/tree/b1b4f05d50d60ea2bf2350b5a287c83499db0540
>         
>         * The LICENSE file for it is here, and it implies that some files may use other licenses: https://github.com/jrl-umi3218/jrl-cmakemodules/blob/b1b4f05d50d60ea2bf2350b5a287c83499db0540/LICENSE
>         * Weird thing is that repo has a `package.xml` file and its license is listed as BSD-2: https://github.com/jrl-umi3218/jrl-cmakemodules/blob/b1b4f05d50d60ea2bf2350b5a287c83499db0540/package.xml#L11
>         * The license for this should be clarified, and all of its licenses should be listed as `<license>` under _your_ package, since you're including it as a submodule
> 
> 
> 
> Overall, the licensing needs to be clarified. I know the original package (eigenpy) does the same thing and has been released already, but I don't think it's a valid reason to do the same here unfortunately.

nim65s:
fix #763
ref https://github.com/ros/rosdistro/pull/45744

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -3,16 +3,109 @@
   <name>jrl-cmakemodules</name>
   <version>0.0.0</version>
   <description>CMake utility toolbox</description>
+  <url type="website">https://github.com/jrl-umi3218/jrl-cmakemodules</url>
 
   <maintainer email="guilhem.saurel@laas.fr">Guilhem Saurel</maintainer>
   <author>Justin Carpentier</author>
   <author>Pierre Gergondet</author>
 
-  <license>BSD-2</license>
-  <url type="website">https://github.com/jrl-umi3218/jrl-cmakemodules</url>
+  <license>GPL-3.0</license>
+  <!-- GPL-3.0 files:
+        - xacro.cmake
+        - compiler.cmake
+        - idlrtc.cmake
+        - idl.cmake
+        - gtest.cmake
+        - debian.cmake
+        - distcheck.cmake
+        - cmake_uninstall.cmake.in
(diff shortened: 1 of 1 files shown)


package.xml
```diff
@@ -3,16 +3,109 @@
...
+        - cpack.cmake
+  -->
 
-  <buildtool_depend>cmake</buildtool_depend>
+  <license>LGPL-3.0</license>
+  <!-- LGPL-3.0 files:
+        - tracy.cmake
+        - find-external/CoinUtils/FindCoinUtils.cmake
+        - find-external/TinyXML/FindTinyXML.cmake
+        - find-external/Ipopt/FindIpopt.cmake
```
ManifoldFR, line 81: Not entirely sure that the Ipopt finder module I wrote is LGPL 3.0. But I did take inspiration from Casadi's own file initially. But no code from there was copied.

PR #771, opened 2025-07-08
Use REPLACE instead of REGEX REPLACE

jorisv:
`string(REGEX REPLACE ...)` should only be used when using a regex. To replace some string, it's better to use `string(REPLACE ...)`.

Fix https://github.com/Simple-Robotics/proxsuite/issues/405


nim65s (approved):
It is better, and it works for me, thanks :)

PR #772, opened 2025-07-22
Don't add dependency to the generated CMake module if the dependency is not found

jorisv:
This PR do the following:
- Don't add a dependency added by `ADD_PROJECT_DEPENDENCY` macro if the dependency is not found in the generated CMake module
- Make package-config cached variable reentrant.
- Remove PACKAGE_EXTRA_MACROS from the INTERNAL CACHE since it's modified by user

Tested on coal and Pinocchio.


package-config.cmake
```diff
@@ -250,7 +250,7 @@ macro(SETUP_PROJECT_PACKAGE_FINALIZE)
     "${${PROJECT_NAME}_PACKAGE_CONFIG_DEPENDENCIES_PROJECTS}"
   )
   set(
-    PACKAGE_CONFIG_DEPENDENCIES_FINDE_EXTERNAL
+    PACKAGE_CONFIG_DEPENDENCIES_FIND_EXTERNAL
```
nim65s: This explain a few things, thanks :D

nim65s (commented):
overall this looks good to me. I'm currently rebuilding things with it to double check.

NB: reviewing the first commit without `?w=1` is impossible 😅

nim65s:
This is breaking proxsuite v0.7.2 configure step with system simde:
```
proxsuite> CMake Error at /nix/store/68xlk52gr0vjyhz3sp2sqzlanndkykcx-jrl-cmakemodules-0-unstable-2025-05-04/share/jrl-cmakemodules/package-config.cmake:102 (find_package):
proxsuite>   By not providing "FindSimde.cmake" in CMAKE_MODULE_PATH this project has
proxsuite>   asked CMake to find a package configuration file provided by "Simde", but
proxsuite>   CMake did not find one.
proxsuite>   Could not find a package configuration file provided by "Simde" with any of
proxsuite>   the following names:
proxsuite>     SimdeConfig.cmake
proxsuite>     simde-config.cmake
proxsuite>   Add the installation prefix of "Simde" to CMAKE_PREFIX_PATH or set
proxsuite>   "Simde_DIR" to a directory containing one of the above files.  If "Simde"
proxsuite>   provides a separate development package or SDK, be sure it has been
proxsuite>   installed.
proxsuite> Call Stack (most recent call first):
proxsuite>   CMakeLists.txt:187 (ADD_PROJECT_DEPENDENCY)

File: package-config.cmake [hunk 2]
```diff
@@ -81,77 +96,91 @@ macro(ADD_PROJECT_DEPENDENCY)
     ""
     ${ARGN}
   )
-  if(PARSED_ARGN_PKG_CONFIG_REQUIRES)
+
+  list(GET PARSED_ARGN_UNPARSED_ARGUMENTS 0 _package_name)
+  if(NOT ${_package_name} IN_LIST PROJECT_PACKAGES_IN_WORKSPACE)
+    find_package(${PARSED_ARGN_UNPARSED_ARGUMENTS})
```
nim65s, line 102: That `find_package` is now before the `set(CMAKE_MODULE_PATH …)`, which is breaking the feature

jorisv: Nice catch. This should be fixed now.

PR #815, opened 2026-05-22
generate pypa metadata

nim65s:
Hi,

This generate PEP 643 `${PYTHON_SITELIB}/${PROJECT_NAME}-${PROJECT_VERSION}.dist-info/METADATA` from PEP 621 pyproject.toml with https://github.com/pypa/pyproject-metadata/

### DIFF
--- a/pypa-metadata.py
+++ b/pypa-metadata.py
@@ -23,6 +23,8 @@ def generate_metadata(pyproject: Path, site_packages: Path):
 
     dist = f"{metadata.name.replace('-', '_')}-{metadata.version}"
     dist_info = site_packages / f"{dist}.dist-info"
+    # This print is not a debugging information but is parsed by the CMake script
+    # to know where to install generated file.
     print(dist_info)
     dist_info.mkdir(parents=True)
 
--- a/pypa-metadata.py
+++ b/pypa-metadata.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+"""
+Generate PEP 643 metadata from PEP 621 pyproject.toml
+
+ref. https://packaging.python.org/en/latest/specifications/recording-installed-packages/
+"""
+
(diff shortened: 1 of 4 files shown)


pypa-metadata.py
```diff
@@ -0,0 +1,35 @@
...
+parser.add_argument("site_packages", type=Path, help="Path to python site-package")
+
+
+def generate_metadata(pyproject: Path, site_packages: Path):
+    with pyproject.open("rb") as f:
+        metadata = StandardMetadata.from_pyproject(load(f))
+
+    dist = f"{metadata.name.replace('-', '_')}-{metadata.version}"
+    dist_info = site_packages / f"{dist}.dist-info"
+    print(dist_info)
```
jorisv, line 28:
```suggestion
    # This print is not a debugging information but is parsed by the CMake script
    # to know where to install generated file.
    print(dist_info)
```

pypa-metadata.py
```diff
@@ -0,0 +1,35 @@
...
+from pyproject_metadata import StandardMetadata
+
+
+parser = ArgumentParser()
+parser.add_argument("pyproject", type=Path, help="Path to source pyproject.toml")
+parser.add_argument("site_packages", type=Path, help="Path to python site-package")
+
+
+def generate_metadata(pyproject: Path, site_packages: Path):
+    with pyproject.open("rb") as f:
```
jorisv, line 21: Did you really need "b" ?

jorisv (changes requested):

pypa-metadata.py
nim65s: the default is text mode: https://docs.python.org/3/library/functions.html#open

and tomllib explicitely wants a binary object: https://docs.python.org/3/library/tomllib.html#tomllib.load

jorisv (approved):

PR #819, opened 2026-06-22
[feature] `jrl_generate_options_markdown_summary()`

ahoarau:
Add a function to generate a summary table in markdown format.
The idea is to generate the summary in the *source* directory (ex: `docs/options.md`), and add the link in the main `README.md` to people can find it.

NOTE: it would be possible to add fake comments in the main `README.md`, and place the summary in between. 
Anything's possible :-)

```cmake
jrl_generate_options_markdown_summary()
```

Example in `coal`:

image


nim65s (approved):
Thanks !

PR #820, opened 2026-06-26
nix: run tests

nim65s:
And fix a few ones

### DIFF
--- a/v2/tests/CMakeLists.txt
+++ b/v2/tests/CMakeLists.txt
@@ -15,7 +15,7 @@ add_test(
         ${CMAKE_COMMAND} -E cat
         "${jrl-cmakemodules_INSTALL_DIR}/share/cmake/jrl-cmakemodules/v2/modules/jrl.cmake"
 )
-set_tests_properties("Check jrl.cmake" PROPERTIES DEPENDS jrl-cmakemodules)
+set_tests_properties("Check jrl.cmake" PROPERTIES DEPENDS jrl-cmakemodules RUN_SERIAL TRUE)
 
 # 3. Run the tests with the local test projects
 add_subdirectory(find_modules)
--- a/v2/tests/python_bindings/pb-library/bindings/boostpy-bindings/CMakeLists.txt
+++ b/v2/tests/python_bindings/pb-library/bindings/boostpy-bindings/CMakeLists.txt
@@ -24,6 +24,7 @@ jrl_python_compile_all(DIRECTORY
 )
 
 jrl_python_compute_install_dir(python_install_dir)
+install(TARGETS pb_pywrap_bp DESTINATION "${python_install_dir}/pb_bp")
 install(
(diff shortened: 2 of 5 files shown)


v2/tests/python_bindings/pb-library/bindings/boostpy-bindings/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ jrl_python_compile_all(DIRECTORY
 )
 
 jrl_python_compute_install_dir(python_install_dir)
+install(TARGETS pb_pywrap_bp DESTINATION ${python_install_dir})
```
ahoarau: The library/wrapper is supposed to be present in the `${CMAKE_BINARY_DIR}/lib/site-packages/pb_bp` directory, installed with the next line.

v2/tests/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ add_test(
         ${CMAKE_COMMAND} -E cat
         "${jrl-cmakemodules_INSTALL_DIR}/share/cmake/jrl-cmakemodules/v2/modules/jrl.cmake"
 )
-set_tests_properties("Check jrl.cmake" PROPERTIES DEPENDS jrl-cmakemodules)
+set_tests_properties("Check jrl.cmake" PROPERTIES DEPENDS jrl-cmakemodules RUN_SERIAL TRUE)
```
ahoarau, line 18: Interesting, but I thought the `DEPENDS` would suffice to enforce building+installing jrl-cmakemodules first ?

v2/tests/python_bindings/pb-library/bindings/boostpy-bindings/CMakeLists.txt
nim65s: ref. commit message:
> tests: pb-library: binaries should be installed as TARGETS, otherwise the `RPATH` rewrite between build and install is not triggered

v2/tests/CMakeLists.txt
nim65s: me too, but I clearly got multiple times an error in that test about that file not existing, even if the file does exists if I look a few seconds afterwards.

I guess it does depend on filesystem implementations: copying a file is not atomic, so maybe the copy takes a little more time afd the first test can end before the file is actually available.

v2/tests/python_bindings/pb-library/bindings/boostpy-bindings/CMakeLists.txt
nim65s: there is a FILES_MATCHING where I removed *.so & *.pyd just below, do you think that is not enough ?

ahoarau: OK interesting. Just need to put the lib in the `pb_bp` dir: 
`install(TARGETS pb_pywrap_bp DESTINATION ${python_install_dir}/pb_bp)`

PR #821, opened 2026-06-26
v2: install cmake export in lib/, not lib64/

nim65s:
because I hate lib64/, but we can discuss : prove me wrong :)

### DIFF
--- a/v2/modules/jrl.cmake
+++ b/v2/modules/jrl.cmake
@@ -2170,8 +2170,7 @@ function(_jrl_export_dependencies)
     if(arg_INSTALL_DESTINATION)
         set(INSTALL_DESTINATION ${arg_INSTALL_DESTINATION})
     else()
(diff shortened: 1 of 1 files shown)


ahoarau:
I still have some reserves about this. Not using `CMAKE_INSTALL_LIBDIR` might break stuff in debian:

image

ahoarau:
In `jrl_configure_default_install_dirs`:

```cmake
    include(GNUInstallDirs)
    # We dont want to use the old lib64, force lib for all platforms (even on 64-bit systems)
    if(CMAKE_INSTALL_LIBDIR STREQUAL "lib64")
        set(CMAKE_INSTALL_LIBDIR "lib" CACHE PATH "Install path for libraries" FORCE)
    endif()
 ```

ahoarau (approved):
It seems to be only limited for CMake export files, so your modification it's fine.

PR #842, opened 2026-07-04
v2(nix): add jrl-release tool

arntanguy:
@nim65s I'm not yet used to package pure python applications, so not sure this is the best nix-way of doing it (it works). 
Also, I'm confused where the "upstream" `jrl-cmakemodule` package is, directly on nixos/nixpkgs? Should I upstream the cmake-parser derivation?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,6 +8,67 @@
     inputs.gepetto.lib.mkFlakoboros inputs (
       { lib, ... }:
       {
+        pyPackages.cmake-parser =
+          {
+            buildPythonPackage,
+            setuptools,
+            setuptools-scm,
+            fetchPypi,
+            attrs,
+            ...
+          }:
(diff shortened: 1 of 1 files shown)


nim65s:
This looks mostly correct, but yes this should go in nixpkgs, and I already have this ready somewhere, sorry for not pushing that earlier

nim65s:
I've opened a draft pull request here: https://github.com/NixOS/nixpkgs/pull/539525

I still need to work a bit on the `TAGFILES` to ensure the documentation of a project correctly links to the documentation of its dependencies.

arntanguy:
Ok feel free to close when it lands on nixpkgs then ;)

I'm interested in the TAGFILES solution, for now cross-linking between Doxygen documentation is done manually in CI, which is not ideal. I'll replace it with Nix' version ;)

PR #843, opened 2026-07-04
jrl_release: Add DebianChangelogVersionExtractor

arntanguy:
@ahoarau In JRL/LIRMM ecosystem, we have debian packaging handled by each repository with a `debian/` folder containing all files expected by `debbuild` tools. We need to ensure that the `debian/changelog` versions remain synced with the project.

Ex: https://github.com/jrl-umi3218/copra/commit/7d4ecb29ec76762521e7991a0adb6ebc09403c57, the first entry was autogenerated by this PR.
```
copra (1.3.3-1debian1) unstable; urgency=medium

  * New release: version 1.3.3

 -- Arnaud TANGUY <arn.tanguy@gmail.com>  Sat, 04 Jul 2026 18:21:00 +0200

copra (1.3.2-1debian1) unstable; urgency=medium

  * Release 1.3.2

 -- Thomas Duvinage <t.duvinage+ppa@outlook.fr>  Wed, 23 Jul 2025 11:01:35 +0900
 ```

This:
- Adds a `DebianChangelogVersionExtractor`
- Gets the full version `1.3.2-1debian1`, the `-1debian1` is a version specific to debian packaging. In practice, I don't think we've ever bumped it, so I think it makes sense to just copy it from the previous version to the new one. If the case is more complex than this, we probably need to be doing some explicit packaging anyways.
- Bump the version number to `1.3.3-1debian1`. The `-1debian1` is a version specific to debian packaging. In practice, I don't think we've ever bumped it, so I think it makes sense to just copy it from the previous version to the new one. If the case is more complex than this, we probably need to be doing some explicit packaging anyways.
- Opinionated choice: The maintainer name and address is obtained from git. That's in line with the other commands that can do commits/tags for us. This has a small implications on tests: I used a monkeypatch to replace the `run_git_command` call with a dummy so that we always get the same user name/mail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+- jrl_release: add `DebianChangelogVersionExtractor` by @arntanguy
+
 ## [2.1.0] - 2026-07-03
 
 - jrl_release: add `--sign-tag` to sign the tag by @ahoarau
--- a/v2/scripts/jrl_release.py
+++ b/v2/scripts/jrl_release.py
@@ -372,6 +372,114 @@ class CMakeListsVersionExtractor(VersionExtractor):
             f.write(content)
 
 
+class DebianChangelogVersionExtractor(VersionExtractor):
+    """Specialized extractor for debian/changelog files.
+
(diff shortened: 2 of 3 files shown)


ahoarau (approved):

ahoarau:
That's great ! I think we need to add a line on the changelog

arntanguy:
Sure, I should take the habit ;)

PR #846, opened 2026-07-06
Fix documentation installation when working outside build directory

jorisv:
When installing with `cmake --install build` we can have the following error:

```bash
ninja: error: loading 'build.ninja': No such file or directory
CMake Error at build/cmake_install.cmake:70 (file):
  file INSTALL cannot find "/tmp/pinocchio/build/doc/doxygen-html": No such
  file or directory.
```
To fix this we use `cmake --build` instead of generator make program so we can provide binary directory path as argument.


jorisv:
Already fixed in another PR

PR #847, opened 2026-07-06
jrl_release: Meta pkg ros support

ahoarau:
THis PR allows discovery of nested packages, like in ROS meta pkgs scenarios.

This adds a bit of complexity in order to ignore walking into ALL the files. So right now we limit to git repositories, avoid folders that starts with ".", mais ignore `.gitignore` files/folders with `git ls-files`, which happens to be very helpful.


nim65s (approved):
lgtm, thanks a lot !

PR #861, opened 2026-08-07
Disable LTO on macos due to an issue on clang feedstock

ahoarau:
Happens on clang 22. The tests pulls the latest version. clangdev have been notified.


ahoarau:
@nim65s This is a workaroung. I try to add `no-changelog`, but it did not have any effect on the Changelog Check CI.

nim65s (commented):
In general, I don't think a change in how conda package one dependency should lead to a change in our sources: we should instead fix that in our pixi config.


LTO seems fine on macos otherwise:
- brew: https://github.com/jrl-umi3218/jrl-cmakemodules/actions/runs/31200866983/job/92940249830#step:7:363
- nix: https://github.com/jrl-umi3218/jrl-cmakemodules/actions/runs/31200865648/job/92940244536#step:6:588


I'm fine with merging this anyway if we don't find a better solution, but if that is only a bug in conda forge, then we can just do nothing and that will be fine too :)

ahoarau:
@nim65s I'm OK with shipping it like this. It's only temporary anyways.

PR #868, opened 2026-08-13
jrl: Add cppad, cppadcg, gmp and mpfr find-modules

ahoarau:
Taken from https://github.com/stack-of-tasks/pinocchio/pull/2923

Important note: 
Initially the `Findcppadcg.cmake` was calling `find_package(cppad REQUIRED)` internally.
The issue is that our `jrl_export_package()` cannot detect transitive find-modules, it only reacts to `jrl_find_package()`.
As we want to keep the find modules jrl-free, they have been edited to be autonomous: `Findcppadcg.cmake` will look for the `cppad::cppad` target internally, making it NOT depend on `Findcppad.cmake`.

An alternative explored was on the user side, with people having to call the required finder before the other:
```
jrl_find_package(cppad REQUIRED)
jrl_find_package(cppadcg REQUIRED)
```
So that the `cppad` find module would get exported. This is not ideal as its difficult to enforce.


ahoarau:
@nim65s I need to invoke your nix magic powers on this one :-) 
We need to add `cppad`, `cppagcg`, `gmp` and `mpfr` if they are available.

nim65s (commented):
I wonder if we should try `pkg_check_modules(… IMPORTED_TARGET)` (or In CMake >= 3.31, `cmake_pkg_config(IMPORT …)`) instead.

Some packages do not export cmake configs, but the .pc file they ship are good.

But probably we should continue using something similar to v1 find-external for now

ahoarau:
I think the objective of the v2 was to drop `pkg-config` completely, but I like the idea. 
Most probably in the future they will ship a cmake config, and drop pkg-config.

nim65s:
probably not gmp or mpfr

ahoarau:
> probably not gmp or mpfr

if gmp and mpfr not possible, I need to add an aditionnal option OFF by default to avoid building this test. Let me know.

nim65s:
I mean I don't think gmp and mpfr will ever switch to CMake

v2/tests/find_modules/CMakeLists.txt
```diff
@@ -24,3 +21,11 @@ endif()
 if(JRL_CMAKEMODULES_ENABLE_TEST_CPPADCG)
     add_cmake_test(NAME Findcppadcg BUILD_TARGET all)
 endif()
+
+if(JRL_CMAKEMODULES_ENABLE_TEST_GMP)
```
nim65s, line 25: couldn't we put those in `if(WIN32)` ?

ahoarau: I'm trying to avoid having the build system decide for anything. 
That said, the windows issue seems to be related to some caching issue, as suitesparse was already pulling GMP and it worked a few weeks ago.

